PR #16, opened 2020-12-22
Added inception model

likholat:


dkurt:
Please add this model and squeezenet from past PR to the README (list of supported models)

modules/mo_pytorch/mo_extensions/front/pytorch/pooling_ext.py
```diff
@@ -78,3 +78,39 @@ def extract(cls, node):
...
+
+class AvgPool2dFrontExtractor(FrontExtractorOp):
+    op = 'AvgPool2d'
+    enabled = True
+
+    @classmethod
+    def extract(cls, node):
+        # Extract pads attribute
+        pads = np.array([node.module.padding, node.module.padding], dtype=np.int64).reshape(1, 2)
+        pads = np.repeat(pads, 2, axis=0)
```
dkurt, line 91: I think that we need to have a unified approach for paddings across layers. Should we do the same as proposed for conv layer?

modules/mo_pytorch/mo_extensions/front/pytorch/slice_ext.py
```diff
@@ -0,0 +1,37 @@
...
+    op = 'StridedSlice'
+    enabled = True
+
+    @classmethod
+    def extract(cls, node: Node):
+
+        attrs = {
+            'begin_mask': [0,1,0,0],
+            'end_mask': [0,1,0,0],
+            'shrink_axis_mask': [0,1,0,0],
```
dkurt, line 31: Please add generic implementation

modules/mo_pytorch/mo_extensions/load/pytorch/loader.py
```diff
@@ -24,6 +24,8 @@
 from torch.autograd import Variable
 
 from extensions.load.loader import Loader
+from extensions.middle.ConvertGroupedStridedSlice import ConvertGroupedStridedSlice
+ConvertGroupedStridedSlice.enabled = False
```
dkurt, line 28: Can you please try to export model to ONNX and convert using MO? Interesting to see if we face similar problem.

likholat: We can get IR model by this way, and it will pass the test

dkurt: Can you please try with the following code?

```python
# Comment
# from extensions.middle.ConvertGroupedStridedSlice import ConvertGroupedStridedSlice
# ConvertGroupedStridedSlice.enabled = False
```

```python
# front/pytorch/slice_ext.py

from extensions.middle.ConvertGroupedStridedSlice import ConvertGroupedStridedSlice

class StridedSliceFrontExtractor(FrontExtractorOp):
    op = 'StridedSlice'
    enabled = True

  def run_after(self):
    return [ConvertGroupedStridedSlice]
```

likholat: The same problem

modules/mo_pytorch/test/run_tests.py
```diff
@@ -80,6 +80,7 @@ def check_torchvision_model(self, model_func, size, threshold=1e-5):
 
         # Run model with OpenVINO and compare outputs
         net = self.ie.read_network('model.xml', 'model.bin')
+
```
dkurt, line 83: extra new line

modules/mo_pytorch/mo_extensions/load/pytorch/hooks.py
```diff
@@ -112,14 +112,28 @@ def __init__(self, value):
...
     def __getitem__(self, n):
+        import numpy as np
+
+        value_ndim = self._value.ndim
+        begin_id = np.zeros(value_ndim, dtype=int)
+        end_id = np.zeros(value_ndim, dtype=int)
+        mask_arr = np.zeros(value_ndim, dtype=int)
+
+        for i in range(len(n)):
+            if isinstance(n[i], int):
```
dkurt, line 123: What other types we can have?

modules/mo_pytorch/mo_extensions/front/pytorch/slice_ext.py
```diff
@@ -17,18 +17,17 @@
...
     def extract(cls, node: Node):
-
+        mask = node.module.mask
         attrs = {
-            'begin_mask': [0,1,0,0],
-            'end_mask': [0,1,0,0],
-            'shrink_axis_mask': [0,1,0,0],
+            'begin_mask': mask,
+            'end_mask': mask,
+            'shrink_axis_mask': mask,
```
dkurt, line 30: Why the same `mask` is used for multiple different attributes?

modules/mo_pytorch/mo_extensions/front/pytorch/common.py
```diff
@@ -0,0 +1,13 @@
+import numpy as np
+
+def get_pads(padding):
```
dkurt, line 3: Will it works for other attributes such strides, kernel_size, dilations?

modules/mo_pytorch/mo_extensions/front/pytorch/slice_ext.py
likholat: `begin_mask[i]` equal to 1 means that the corresponding dimension of the begin input is ignored and the real beginning of the tensor is used along corresponding dimension. The same for `end_mask`. This masks will be flipped. 
And if `shrink_axis_mask[i]` is 1, the dimension on the i-th position is deleted, but we should remove the dimension by which we make the slice. 
So we should change `begin_mask`, `end_mask` and `shrink_axis_mask` for the dimension we change.

modules/mo_pytorch/mo_extensions/load/pytorch/hooks.py
likholat: tuple

modules/mo_pytorch/mo_extensions/front/pytorch/slice_ext.py
likholat: begin_mask[i] equal to 1 means that the corresponding dimension of the begin input is ignored and the real beginning of the tensor is used along corresponding dimension. The same for end_mask. This masks will be flipped.
And if shrink_axis_mask[i] is 1, the dimension on the i-th position is deleted, but we should remove the dimension by which we make the slice.
So we should change begin_mask, end_mask and shrink_axis_mask for the dimension we change.

modules/mo_pytorch/mo_extensions/front/pytorch/common.py
likholat: tuple

modules/mo_pytorch/mo_extensions/load/pytorch/hooks.py
dkurt: Do we know how to manage tuple in this way? There is no `else` here

modules/mo_pytorch/mo_extensions/front/pytorch/slice_ext.py
dkurt: Yes, but what about `a[:, 2:, :, :]`, in example.  This way
```
begin_mask: [0, 1, 0, 0]
end_mask: [0, 0, 0, 0]
```

modules/mo_pytorch/mo_extensions/front/pytorch/common.py
likholat: No, it won't. We have doubled dimension for each value of padding.

modules/mo_pytorch/mo_extensions/load/pytorch/hooks.py
```diff
@@ -96,11 +96,73 @@ class Add(nn.Module):
...
+            class Add(nn.Module):
+                def __init__(self, value):
+                    super().__init__()
+                    self.register_buffer('add', torch.tensor(value))
+
+            res = self._value + a
+            return forward_hook(Add(a), (self,), res)
+
+    def __getitem__(self, n):
+        import numpy as np
```
dkurt, line 115: unused import

modules/mo_pytorch/mo_extensions/load/pytorch/hooks.py
```diff
@@ -96,11 +96,73 @@ class Add(nn.Module):
...
+
+        value_ndim = self._value.ndim
```
dkurt, line 117: unused variable

modules/mo_pytorch/mo_extensions/load/pytorch/hooks.py
```diff
@@ -96,11 +96,73 @@ class Add(nn.Module):
...
+        elif isinstance(a, float):
+            class Add(nn.Module):
+                def __init__(self, value):
+                    super().__init__()
+                    self.register_buffer('add', torch.tensor(value))
+
+            res = self._value + a
+            return forward_hook(Add(a), (self,), res)
+
+    def __getitem__(self, n):
```
dkurt, line 114: Please rename `n` to something meaningful. In example, `key` (see https://docs.python.org/3/reference/datamodel.html#object.__getitem__)

modules/mo_pytorch/mo_extensions/load/pytorch/hooks.py
```diff
@@ -96,11 +96,73 @@ class Add(nn.Module):
...
+                begin_mask.append(1)
+                end_mask.append(1)
+
+            elif isinstance(item, slice):
+                if item.start is not None:
+                    begin_id.append(item.start)
+                    begin_mask.append(1)
+                else:
+                    begin_id.append(0)
+                    begin_mask.append(0)
```
dkurt, line 139:
```python
begin_id.append(item.start if item.start else 0)
begin_mask.append(1 if item.start else 0)
```

modules/mo_pytorch/mo_extensions/load/pytorch/hooks.py
```diff
@@ -96,11 +96,73 @@ class Add(nn.Module):
...
+
+                if item.stop is not None:
+                    end_id.append(item.stop)
+                    end_mask.append(1)
+                else:
+                    end_id.append(0)
+                    end_mask.append(0)
```
dkurt, line 146: Same as above

modules/mo_pytorch/mo_extensions/load/pytorch/hooks.py
```diff
@@ -96,11 +96,73 @@ class Add(nn.Module):
...
+
+                if (end_id[-1] - begin_id[-1] != 1):
+                    shrink_axis_mask.append(0)
+                else:
+                    shrink_axis_mask[i].append(1)
```
dkurt, line 151: `[i]`?

modules/mo_pytorch/mo_extensions/front/pytorch/conv_ext.py
```diff
@@ -21,23 +21,21 @@
...
     op = 'Conv2d'
     enabled = True
 
     @classmethod
     def extract(cls, node):
-        # Extract pads attribute
-        pads = np.array(node.module.padding, dtype=np.int64).reshape(1, 2)
-        pads = np.repeat(pads, 2, axis=0)
-        final_pads = np.array([[0, 0], [0, 0], *pads], dtype=np.int64)
+        final_pads = get_pads(node.module.padding)
```
dkurt, line 32: It's better to accept nn.Moduleas input:
```python
final_pads = get_pads(node.module)
```
Internally, get `.padding` attribute

modules/mo_pytorch/README.md
```diff
@@ -36,3 +36,5 @@ This module let you convert PyTorch models directly to OpenVINO IR without using
 * `torchvision.models.resnet18`
 * `torchvision.models.segmentation.deeplabv3_resnet50`
 * `Detectron2 RetinaNet`
+* `torchvision.models.squeezenet1_1`
+* `torchvision.models.inception_v3`
```
dkurt, line 40: Please change the order:
```
torchvision.models.alexnet
torchvision.models.resnet18
torchvision.models.squeezenet1_1
torchvision.models.inception_v3
torchvision.models.segmentation.deeplabv3_resnet50
Detectron2 RetinaNet
```

modules/mo_pytorch/mo_extensions/load/pytorch/loader.py
dkurt: Please add a note:
```python
from extensions.middle.ConvertGroupedStridedSlice import ConvertGroupedStridedSlice
# TODO: fix this bug
ConvertGroupedStridedSlice.enabled = False
```

dkurt (approved):
Well done!

PR #33, opened 2021-02-10
Supported Proposal via ngraph::reference

l-bat:


modules/arm_plugin/src/arm_converter/arm_converter_proposal.cpp
```diff
@@ -0,0 +1,52 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+#include <details/ie_exception.hpp>
+
+#include "arm_converter/arm_converter.hpp"
+#include <ngraph/runtime/reference/proposal.hpp>
+
+namespace ArmPlugin {
+template<> Converter::Conversion::Ptr Converter::Convert(const opset::Proposal& node) {
+    auto make = [&] (auto refFunction) {
```
apankratovantonp, line 11: `MakeConversion` can be used here directly without `make` wrapper as it is called just once

apankratovantonp (approved):

PR #44, opened 2021-02-16
Supported OneHot via ngraph::reference

l-bat:


modules/arm_plugin/src/arm_converter/arm_converter_one_hot.cpp
```diff
@@ -0,0 +1,89 @@
+// Copyright (C) 2020-2021 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+
+#include <details/ie_exception.hpp>
+
+#include "arm_converter/arm_converter.hpp"
+#include <ngraph/runtime/reference/one_hot.hpp>
+
+namespace ArmPlugin {
+template <typename INDICES_TYPE, typename OUTPUT_TYPE>
```
apankratovantonp, line 10: Do not use UPPER_CASE for template parameters. It is commonly used for MACRO. Just use: `INDICES_TYPE` -> `Indices`

l-bat: Okay) I copied it from nGraph

apankratovantonp (approved):

PR #48, opened 2021-02-18
Updated dockerfiles and script for docker-based compilation of ARM plugin

terfendail:


terfendail:
@alvoron, @apankratovantonp, @l-bat please review

modules/arm_plugin/armplg_build.sh
```diff
@@ -0,0 +1,212 @@
...
+make -j$BUILD_JOBS && \
+make install && \
+echo export OpenCV_DIR=\$INSTALLDIR/opencv/cmake > $STAGING_DIR/opencv/setupvars.sh && \
+echo export LD_LIBRARY_PATH=\$INSTALLDIR/opencv/lib:\$LD_LIBRARY_PATH >> $STAGING_DIR/opencv/setupvars.sh && \
+cd $DEV_HOME || fail 11 "OpenCV build failed. Stopping"
+
+#Build OpenVINO
+mkdir -p $OPENVINO_HOME/build && \
+cd $OPENVINO_HOME/build && \
+cmake -DOpenCV_DIR=$STAGING_DIR/opencv/cmake -DENABLE_OPENCV=OFF \
```
l-bat, line 109: Why `-DENABLE_OPENCV` is `OFF`?

modules/arm_plugin/armplg_build.sh
```diff
@@ -0,0 +1,212 @@
...
+            echo "Removing existing sources..."
+            rm -rf $1 || fail 1 "Failed to remove. Stopping"
+            echo "Cloning master branch only..."
+            git lfs clone --recurse-submodules --single-branch --branch=master $2 $1 || fail 3 "Failed to clone $TEST_DIR_NAME. Stopping"
+        elif [ -d $1/build ]; then
+            echo "Build directory detected at $1"
+            if [ "$UPDATE_SOURCES" = "clean" ]; then
+                echo "Cleanup of previous build requested"
+                echo "Removing previous build results..."
+                rm -rf $2/build || fail 2 "Failed to cleanup. Stopping"
```
alvoron, line 54: Shouldn't we use $1 here?

modules/arm_plugin/armplg_build.sh
```diff
@@ -0,0 +1,212 @@
...
+#cleanup package destination folder
+[ -e $STAGING_DIR ] && rm -rf $STAGING_DIR
+mkdir -p $STAGING_DIR
+
+#Build OpenCV
+mkdir -p $OPENCV_HOME/build && \
+cd $OPENCV_HOME/build && \
+PYTHONVER=`ls /usr/include | grep "python3[^m]*$"` && \
+cmake -DCMAKE_BUILD_TYPE=$BUILD_TYPE -DBUILD_LIST=imgcodecs,videoio,highgui,python3 \
+      -DBUILD_opencv_python2=OFF -DBUILD_opencv_python3=ON -DOPENCV_SKIP_PYTHON_LOADER=ON \
```
alvoron, line 77: Why do we add python dependency here?

modules/arm_plugin/armplg_build.sh
terfendail: It's an OpenVINO legacy. Actually the flag should be named something like DOWNLOAD_PREBUILD_OPENCV

modules/arm_plugin/armplg_build.sh
terfendail: fixed

modules/arm_plugin/armplg_build.sh
terfendail: OpenVINO python samples require OpenCV as well. So it make sense either enable or disable python support for both. However IMO python is quite popular and worth adding to the generated package.

modules/arm_plugin/README.md
```diff
@@ -25,18 +25,35 @@ git clone --recurse-submodules --single-branch --branch=master https://github.co
...
+>
+> *BUILD_JOBS* control number of threads for simultaneous compilation
+>
+> *BUILD_TYPE* control Debug/Release configuration (Release by default)
+>
+> *UPDATE_SOURCES* control sources retrievement
+>  - *clean*  - don't reload sources if already loaded, just clean build folders(default)
+>  - *reload* - delete all loaded sources and retrieve them again
+>  - *check*  - don't reload or cleanup sources
+6. Extract the archive to `build` directory
```
alvoron, line 56: Probably it would be helpful to add "on target ARM platform" to the end? 
Just to highlight the fact that it's cross-compilation and no need to run binaries on build machine.

terfendail: It's the reasonable point, however it doesn't make sense to call destination directory `build` for target platform.
Also shell command for deployment to target platform probably would be different and platform dependent.
So IMO this step should be just: "6. Extract the archive to target ARM platform"

alvoron: Agree. We may delete `mkdir` command and add `--one-top-level` to tar command, so all content is extracted into `OV_ARM_package`.

alvoron (approved):

modules/arm_plugin/README.md
terfendail: Fixed

l-bat (approved):
👍

apankratovantonp (approved):

modules/arm_plugin/Dockerfile.RPi64
```diff
@@ -4,68 +4,68 @@
...
+    echo deb [arch=amd64] http://archive.ubuntu.com/ubuntu/ bionic-updates universe >> /etc/apt/sources.list && \
+    echo deb [arch=amd64] http://archive.ubuntu.com/ubuntu/ bionic multiverse >> /etc/apt/sources.list && \
+    echo deb [arch=amd64] http://archive.ubuntu.com/ubuntu/ bionic-updates multiverse >> /etc/apt/sources.list && \
+    echo deb [arch=amd64] http://archive.ubuntu.com/ubuntu/ bionic-backports main restricted universe multiverse >> /etc/apt/sources.list && \
+    echo deb [arch=amd64] http://security.ubuntu.com/ubuntu/ bionic-security main restricted >> /etc/apt/sources.list && \
+    echo deb [arch=amd64] http://security.ubuntu.com/ubuntu/ bionic-security universe >> /etc/apt/sources.list && \
+    echo deb [arch=amd64] http://security.ubuntu.com/ubuntu/ bionic-security multiverse >> /etc/apt/sources.list && \
+    echo deb [arch=arm64] http://ports.ubuntu.com/ubuntu-ports/ bionic main >> /etc/apt/sources.list && \
+    echo deb [arch=arm64] http://ports.ubuntu.com/ubuntu-ports/ bionic-updates main >> /etc/apt/sources.list && \
+    echo deb [arch=arm64] http://ports.ubuntu.com/ubuntu-ports/ bionic-security main >> /etc/apt/sources.list && \
```
dkurt, line 24:
```
sed -i -E 's|^deb ([^ ]+) (.*)$|deb [arch=amd64] \1 \2\ndeb [arch=arm64] http://ports.ubuntu.com/ubuntu-ports/ \2|' /etc/apt/sources.list
```

terfendail: amd64 and arm64 repositories aren't mirrored. Some amd64 repositories aren't supported for arm64.(If I get the meaning of sed ... right)

dkurt: @terfendail, this line creates a copy of every entry in `/etc/apt/sources.list` with `[arch=arm64]`. So we can avoid explicit names.

terfendail: I mean original `sources.list` consist of 8 records, but only 3 of them are valid for `[arch=arm64]`

PR #52, opened 2021-02-24
OneHot convertor update

IvanNovoselov:
Update arm_converter_one_hot due to revision of OneHot reference implementation (more details could be found in this [PR](https://github.com/openvinotoolkit/openvino/pull/4330))
Ticket: 37522

### DIFF
--- a/modules/arm_plugin/src/arm_converter/arm_converter_one_hot.cpp
+++ b/modules/arm_plugin/src/arm_converter/arm_converter_one_hot.cpp
@@ -5,23 +5,23 @@
 #include "arm_converter/arm_converter.hpp"
 #include <ngraph/runtime/reference/one_hot.hpp>
 
-template <typename INPUT_TYPE>
-void wrap_one_hot(const INPUT_TYPE* indices,
-             const ngraph::Shape& indices_shape,
-             char* out,
-             const size_t out_elem_size,
-             const size_t depth,
-             const int64_t one_hot_axis,
-             const char* on_value,
-             const char* off_value) {
-    ngraph::runtime::reference::one_hot<INPUT_TYPE>(indices,
-                                            indices_shape,
-                                            out,
-                                            out_elem_size,
-                                            depth,
-                                            one_hot_axis,
-                                            on_value,
-                                            off_value);
+template <typename InputType, typename OutputType>
+void wrap_one_hot(const InputType* indices,
+                  const ngraph::Shape& indices_shape,
+                  OutputType* out,
+                  const size_t out_elem_size,
+                  const std::int64_t depth,
+                  const std::int64_t one_hot_axis,
+                  const OutputType* on_value,
+                  const OutputType* off_value) {
+    ngraph::runtime::reference::one_hot<InputType>(indices,
+                                                   indices_shape,
+                                                   reinterpret_cast<char*>(out),
+                                                   out_elem_size,
+                                                   static_cast<size_t>(depth),
+                                                   one_hot_axis,
+                                                   reinterpret_cast<const char*>(on_value),
+                                                   reinterpret_cast<const char*>(off_value));
 }
 
 namespace ArmPlugin {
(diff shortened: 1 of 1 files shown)


modules/arm_plugin/src/arm_converter/arm_converter_one_hot.cpp
```diff
@@ -2,88 +2,44 @@
 // SPDX-License-Identifier: Apache-2.0
 
 #include <details/ie_exception.hpp>
-
 #include "arm_converter/arm_converter.hpp"
 #include <ngraph/runtime/reference/one_hot.hpp>
 
 namespace ArmPlugin {
-template <typename Indices, typename OutputType>
-void wrap_one_hot(const Indices* arg,
```
l-bat, line 11: Why did you delete wrapper?

IvanNovoselov: I thought we don't need a function that does nothing but calls another function, so I fed ngraph::runtime::reference::xxx directly to MakeConversion(), like it is done [here](https://github.com/openvinotoolkit/openvino_contrib/blob/de5601166c0a5820fbf6f4644e7e3f74d921dcf8/modules/arm_plugin/src/arm_converter/arm_converter_cumsum.cpp#L18) or [here](https://github.com/openvinotoolkit/openvino_contrib/blob/de5601166c0a5820fbf6f4644e7e3f74d921dcf8/modules/arm_plugin/src/arm_converter/arm_converter_conv_backprop_data.cpp#L32) for example. But your question makes me think that I'm not getting it right. 
So could you please explain, why do we need the wrapper? To perform checks, like you mentioned in [this comment](https://github.com/openvinotoolkit/openvino_contrib/pull/52#discussion_r581815735)?

l-bat: We use wrappers to work with inconstant nodes. Instead of extraction a depth value at the graph construction stage, we pass an input node. The value will be extracted at the infer stage. We also need wrappers if `ngraph:: reference` has `char*` input value.

IvanNovoselov: Thank you! Will bring the wrapper back.

modules/arm_plugin/src/arm_converter/arm_converter_one_hot.cpp
```diff
@@ -2,88 +2,44 @@
...
-                                        one_hot_axis,
-                                        on_values[0],
-                                        off_values[0]);
-}
+    template<> Converter::Conversion::Ptr Converter::Convert(const opset::OneHot& node) {
+        const auto& out_shape = node.get_output_shape(0);
+        const std::int64_t axis = node.get_axis();
+        if (axis < 0 || axis >= out_shape.size())
+            THROW_IE_EXCEPTION << "Invalid axis value. Expected in [0, " << out_shape.size() <<
+                                "), got " << axis;
```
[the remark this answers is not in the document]
IvanNovoselov: Negative axis handling is added.

modules/arm_plugin/src/arm_converter/arm_converter_one_hot.cpp
```diff
@@ -2,88 +2,44 @@
...
+        const auto& depth_const = std::dynamic_pointer_cast<opset::Constant>(node.input_value(1).get_node_shared_ptr());
```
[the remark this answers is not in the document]
IvanNovoselov: Check is added.

IvanNovoselov:
@l-bat please note that only int32 and int64 input types are supported by OneHot according to the updated [specs](https://github.com/openvinotoolkit/openvino/blob/master/docs/ops/sequence/OneHot_1.md). Is it Ok that the convertor supports [different types](https://github.com/IvanNovoselov/openvino_contrib/blob/9a176638bad3e025a668c0f672028c8e1a9fb3b9/modules/arm_plugin/src/arm_converter/arm_converter_one_hot.cpp#L63-L100)?

l-bat:
> @l-bat please note that only int32 and int64 input types are supported by OneHot according to the updated [specs](https://github.com/openvinotoolkit/openvino/blob/master/docs/ops/sequence/OneHot_1.md). Is it Ok that the convertor supports [different types](https://github.com/IvanNovoselov/openvino_contrib/blob/9a176638bad3e025a668c0f672028c8e1a9fb3b9/modules/arm_plugin/src/arm_converter/arm_converter_one_hot.cpp#L63-L100)?

Let's add a check that the input types are `int32 `or `int64`, and remove the handling for the other types.

modules/arm_plugin/src/arm_converter/arm_converter_one_hot.cpp
```diff
@@ -2,88 +2,111 @@
...
-                  const OutputType* off_values) {
-    ngraph::runtime::reference::one_hot(arg,
-                                        out,
-                                        in_shape,
-                                        out_shape,
-                                        one_hot_axis,
-                                        on_values[0],
-                                        off_values[0]);
+                  const size_t out_elem_size,
+                  const std::int64_t depth,
```
l-bat, line 13:
```suggestion
                  const InputType* depth,
```

and extract the depth value below. With these changes, we will be able to work with inconstant depth.

modules/arm_plugin/src/arm_converter/arm_converter_one_hot.cpp
```diff
@@ -2,88 +2,111 @@
...
+                  const std::int64_t one_hot_axis,
+                  const OutputType* on_value,
+                  const OutputType* off_value) {
+    ngraph::runtime::reference::one_hot<InputType>(indices,
+                                                   indices_shape,
+                                                   reinterpret_cast<char*>(out),
+                                                   out_elem_size,
+                                                   static_cast<size_t>(depth),
```
l-bat, line 21:
```suggestion
                                                   static_cast<size_t>(depth[0]),
```

modules/arm_plugin/src/arm_converter/arm_converter_one_hot.cpp
```diff
@@ -2,88 +2,111 @@
...
+        if (depth != out_shape[axis])
+            THROW_IE_EXCEPTION << "Incompatible depth and axis values.";
+
+        auto make = [&] (auto refFunction) {
+            return MakeConversion(refFunction,
+                                  node.input(0),
+                                  ind_shape,
+                                  node.output(0),
+                                  node.get_output_element_type(0).size(),
+                                  depth,
```
l-bat, line 56:
```suggestion
                                  node.input(1),
```

IvanNovoselov: Thank you! Your suggestion was incorporated in the latest commit.

modules/arm_plugin/src/arm_converter/arm_converter_one_hot.cpp
IvanNovoselov: Thank you! Your suggestion was incorporated in the latest commit.

modules/arm_plugin/src/arm_converter/arm_converter_one_hot.cpp
IvanNovoselov: Thank you! Your suggestion was incorporated in the latest commit.

l-bat (approved):
Thank you 👍

IvanNovoselov:
@l-bat thank you! Also, could you please delete the [OneHot_revision](https://github.com/openvinotoolkit/openvino_contrib/tree/OneHot_revision) branch? I pushed it by mistake, but have no permission to delete it.

l-bat:
> @l-bat thank you! Also, could you please delete the [OneHot_revision](https://github.com/openvinotoolkit/openvino_contrib/tree/OneHot_revision) branch? I pushed it by mistake, but have no permission to delete it.

Done

IvanNovoselov:
@l-bat some single-layer tests fail [Link to Jenkins](https://openvino-ci.intel.com/job/private-ci/job/ie/job/ie-tests-linux-debian_9_arm-cpu_arm/6545/testReport/). It appears to me that the problem is in the depth and indices type mismatch ([failing SLT source](https://github.com/openvinotoolkit/openvino/blob/799f34238ede58788d6f81f742d8260a992bb0c8/inference-engine/tests/functional/plugin/cpu/shared_tests_instances/single_layer_tests/one_hot.cpp#L45-L99)), so the types of depth and indices must be both integral but could still be different. Shall I revert the last commit and perform checks on depth in the converter?

l-bat:
> @l-bat some single-layer tests fail [Link to Jenkins](https://openvino-ci.intel.com/job/private-ci/job/ie/job/ie-tests-linux-debian_9_arm-cpu_arm/6545/testReport/). It appears to me that the problem is in the depth and indices type mismatch ([failing SLT source](https://github.com/openvinotoolkit/openvino/blob/799f34238ede58788d6f81f742d8260a992bb0c8/inference-engine/tests/functional/plugin/cpu/shared_tests_instances/single_layer_tests/one_hot.cpp#L45-L99)), so the types of depth and indices must be both integral but could still be different. Shall I revert the last commit and perform checks on depth in the converter?

No, fix the wrapper (add new template argument)

IvanNovoselov:
> > @l-bat some single-layer tests fail [Link to Jenkins](https://openvino-ci.intel.com/job/private-ci/job/ie/job/ie-tests-linux-debian_9_arm-cpu_arm/6545/testReport/). It appears to me that the problem is in the depth and indices type mismatch ([failing SLT source](https://github.com/openvinotoolkit/openvino/blob/799f34238ede58788d6f81f742d8260a992bb0c8/inference-engine/tests/functional/plugin/cpu/shared_tests_instances/single_layer_tests/one_hot.cpp#L45-L99)), so the types of depth and indices must be both integral but could still be different. Shall I revert the last commit and perform checks on depth in the converter?
> 
> No, fix the wrapper (add new template argument)

Implemented. Now the tests pass without fails ([latest tests](https://openvino-ci.intel.com/job/private-ci/job/ie/job/ie-tests-linux-debian_9_arm-cpu_arm/6617/testReport/)).

azhogov:
CI passed: https://gitlab-icv.inn.intel.com/inference-engine/product-configs/merge_requests/1430

PR #54, opened 2021-02-25
Build OMZ demos with ARM plugin

alvoron:


ilya-lavrenov (approved):

apankratovantonp (approved):

l-bat (approved):

modules/arm_plugin/armplg_build.sh
```diff
@@ -1,14 +1,18 @@
 #!/bin/sh
 
+set -x
```
terfendail, line 3: Do we need shell debugging enabled during execution of the script?

terfendail (approved):

modules/arm_plugin/armplg_build.sh
alvoron: I think it's useful and it doesn't bring many logs in the same time.
Do you prefer not to use it?

terfendail: Both options are fine for me since there anyway a lot of logs produced by the script. Just to ensure the change was intended.

modules/arm_plugin/armplg_build.sh
```diff
@@ -33,23 +33,23 @@ checkSrcTree()
 {
     [ $# -lt 2 ] && fail
 
-    TEST_DIR_NAME=directory
+    BRANCH_NAME=master
     if [ $# -gt 2 ]; then
-        TEST_DIR_NAME="$3 directory"
```
terfendail, line 34: What do you think about making third parameter mandatory?

alvoron: Agree, fixed

PR #57, opened 2021-03-02
Switched to opset6 

l-bat:


alvoron:
Do we need to update arm_opset.md ?

apankratovantonp (approved):

PR #59, opened 2021-03-02
Update arm opset

l-bat:


alvoron:
grep finds `opset5` on lines 3 and 5 of md file + on lines 7 and 12 of `ngraph_opset.hpp.
Do we need to update them as well?

alvoron (approved):

PR #60, opened 2021-03-02
Added change to support gelu-7 in reference implementations

lazarevevgeny:
A new version of the Gelu operation will be introduced https://github.com/openvinotoolkit/openvino/pull/4497. Need to change ARM plugin to properly use reference implementation.

product-config MR 1444


l-bat (approved):
👍

PR #63, opened 2021-03-03
Fix Klocwork issues in Java

dkurt:


modules/java_api/samples/ArgumentParser.java
```diff
@@ -25,24 +25,19 @@ private void printHelp() {
...
+                    input.put(arg, args[++i]);
                 } else {
-                    if (description.containsKey(arg)) {
-                        input.put(arg, args[++i]);
-                    } else {
-                        System.out.println("Non-existent key: '" + arg + "'");
-                        System.exit(0);
-                    }
+                    System.out.println("Non-existent key: '" + arg + "'");
+                    System.exit(0);
```
likholat, line 38: But what if we run:
```
java -cp "benchmark_app.jar" Main -m model.xml -d
```

modules/java_api/samples/benchmark_app/Main.java
```diff
@@ -91,7 +94,7 @@ static Blob blobRandomByte(TensorDesc tDesc) {
         }
 
         byte[] buff = new byte[size];
-        Random rand = new Random();
+        SecureRandom rand = new SecureRandom();
```
likholat, line 97: Why do we need `SecureRandom` here?

dkurt: Static code analyzer stays that Random is not secure :)

modules/java_api/samples/ArgumentParser.java
dkurt: Good point. Fixed

alvoron (approved):

PR #65, opened 2021-03-04
Updated arm plugin build script

terfendail:
Updated arm plugin build script source branches to specific release version

### DIFF
--- a/modules/arm_plugin/armplg_build.sh
+++ b/modules/arm_plugin/armplg_build.sh
@@ -35,7 +35,7 @@ cloneSrcTree()
     shift
     SRCURL=$1
     shift
-    while [ "$#" -gt "0" ]; do
+    while [ $# -gt 0 ]; do
         git lfs clone --recurse-submodules --shallow-submodules --depth 1 --branch=$1 $SRCURL $DESTDIR && return 0
         shift
     done
--- a/modules/arm_plugin/armplg_build.sh
+++ b/modules/arm_plugin/armplg_build.sh
@@ -29,6 +29,19 @@ fail()
     exit $retval
 }
 
+cloneSrcTree()
+{
+    DESTDIR=$1
+    shift
+    SRCURL=$1
+    shift
+    while [ "$#" -gt "0" ]; do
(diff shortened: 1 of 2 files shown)


modules/arm_plugin/armplg_build.sh
```diff
@@ -61,11 +61,11 @@ checkSrcTree()
 
 
 #Prepare sources
-checkSrcTree $OPENCV_HOME https://github.com/opencv/opencv.git master
-checkSrcTree $OPENVINO_HOME https://github.com/openvinotoolkit/openvino.git master
-checkSrcTree $OPENVINO_CONTRIB https://github.com/openvinotoolkit/openvino_contrib.git master
+checkSrcTree $OPENCV_HOME https://github.com/opencv/opencv.git 4.5.1-openvino
```
alvoron, line 64: Ideally, we need to use `4.5.2-openvino` tag here but it doesn't exist yet. I think, we need to keep `master` here.

Another though t came into my mind: probably we may add a functionality to specify 2 tags\branches? If the 1st tag\branch is not found the 2nd one (fallback) is used. In this case we will use the pair `4.5.2-openvino` - `master` for OpenCV and the pair `2021.3` - `release` for OMZ.

terfendail: Updated. The list of branches is unlimited now. All branches will be checked in a sequence until existing will be found.
I think the change could even be ported to master branch with a sequence of next tag/next release branch/master

alvoron (approved):

modules/arm_plugin/armplg_build.sh
```diff
@@ -29,14 +29,27 @@ fail()
     exit $retval
 }
 
+cloneSrcTree()
+{
+    DESTDIR=$1
+    shift
+    SRCURL=$1
+    shift
+    while [ "$#" -gt "0" ]; do
```
alvoron, line 38: Minor comment: do we need to convert number to string here?

terfendail: It's not necessary here. That's due to my habit to armor everything with parentheses, as a safer approach.
Fixed

PR #66, opened 2021-03-04
Enable tests with strided slice

likholat:
StridedSlice was fixed: https://github.com/openvinotoolkit/openvino/pull/4545


dkurt (approved):
Thanks!

PR #105, opened 2021-03-25
Switch Gather to v1

l-bat:
Relates to openvinotoolkit/openvino#4885


ilya-lavrenov (approved):

apankratovantonp:
OV version of gather was increased? And we do not support this new version?

PR #107, opened 2021-03-29
Disable fusing: convolution with activation

l-bat:


apankratovantonp:
I suppose we should make this configurable by user.

l-bat:
> I suppose we should make this configurable by user.

This fusion is useless for all topologies

apankratovantonp (approved):

PR #120, opened 2021-04-13
Added CTCDecoderSeqLen reference implementation

terfendail:


modules/arm_plugin/src/arm_converter/arm_converter_ctc_greedy_decoder_seq_len.cpp
```diff
@@ -0,0 +1,163 @@
...
+                                    node.input(1),
+                                    node.input(2),
+                                    node.output(0),
+                                    node.output(1),
+                                    node.get_input_shape(0),
+                                    node.get_output_shape(0),
+                                    node.get_merge_repeated());
+    };
+
+    switch (node.get_input_element_type(0)) {
```
l-bat, line 22: It would be nice to replace `switch` with `CallSwitch` as in 
https://github.com/openvinotoolkit/openvino_contrib/blob/master/modules/arm_plugin/src/arm_converter/arm_converter_one_hot.cpp#L56

l-bat (approved):
Looks good for me. Thanks!

PR #126, opened 2021-04-28
Added support for opset7::Gather

terfendail:
Depends on [OpenVINO PR#5373](https://github.com/openvinotoolkit/openvino/pull/5373)

Test could be extended to negative batch index since implemented in reference

### DIFF
--- a/modules/arm_plugin/src/arm_converter/arm_converter_gather.cpp
+++ b/modules/arm_plugin/src/arm_converter/arm_converter_gather.cpp
@@ -8,6 +8,25 @@
 #include "arm_converter/arm_converter.hpp"
 
 namespace ArmPlugin {
+template <> Converter::Conversion::Ptr Converter::Convert(const opset::Gather& node) {
+    auto make = [&] (auto refFunction) {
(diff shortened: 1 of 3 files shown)


modules/arm_plugin/src/arm_converter/arm_converter_gather.cpp
```diff
@@ -8,6 +8,25 @@
 #include "arm_converter/arm_converter.hpp"
 
 namespace ArmPlugin {
+template <> Converter::Conversion::Ptr Converter::Convert(const opset::Gather& node) {
```
l-bat, line 11: I think we can create an transformation opset::Gather to opset::ArmGather for some cases to infer it via ACL

terfendail: As far as I could understand there already is [one](https://github.com/openvinotoolkit/openvino_contrib/blob/3e516d960d11f69a67dd30e87a5f37c2a3e7e94d/modules/arm_plugin/src/transformations/convert_gather_arm.cpp#L15)

alvoron (approved):

PR #134, opened 2021-05-13
Fixed deprecation warnings

apankratovantonp:
NMS conversion do not use deprecated `ngraph::Tensor::get_name`

### DIFF
--- a/modules/arm_plugin/src/arm_converter/arm_converter.hpp
+++ b/modules/arm_plugin/src/arm_converter/arm_converter.hpp
@@ -63,12 +63,17 @@ struct Argument<arm_compute::ITensor*> {
     arm_compute::Tensor     _notPaddedTensor;
 };
 
+struct HostTensors {
+    ngraph::HostTensorVector _hosts;
+    const ngraph::Node*      _node = nullptr;
+};
+
 template<>
-struct Argument<ngraph::HostTensorVector> {
-    Argument(const ngraph::HostTensorVector& hosts, const std::vector<arm_compute::ITensor*>& tensors, ArgumentType type) :
+struct Argument<HostTensors> {
+    Argument(const HostTensors& hosts, const std::vector<arm_compute::ITensor*>& tensors, ArgumentType type) :
         _type{type},
         _tensors(tensors),
-        _hosts{hosts} {
+        _hosts{hosts._hosts} {
         _notPaddedTensors.resize(_tensors.size());
         for (size_t i = 0; i < _tensors.size(); i++) {
             if (_tensors[i]->info()->has_padding()) {
@@ -227,8 +232,8 @@ struct Converter {
                 }
             }
 
-            void StartArgumentLifeTime(Argument<ngraph::HostTensorVector>& hostsArgument) {
-                for (size_t i = 0; i < hostsArgument._hosts.size(); i++) {
+            void StartArgumentLifeTime(Argument<HostTensors>& hostsArgument) {
+                for (std::size_t i = 0; i < hostsArgument._hosts.size(); i++) {
                     if (hostsArgument._tensors[i]->info()->has_padding()) {
                         _memoryGroup.manage(&(hostsArgument._notPaddedTensors[i]));
                     }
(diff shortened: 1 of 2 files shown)


modules/arm_plugin/src/arm_converter/arm_converter.hpp
```diff
@@ -63,12 +63,17 @@ struct Argument<arm_compute::ITensor*> {
     arm_compute::Tensor     _notPaddedTensor;
 };
 
+struct HostTensors {
+    ngraph::HostTensorVector _hosts;
+    const ngraph::Node*      _node = nullptr;
```
ilyachur, line 68: Why do you need to have node raw pointer here?

apankratovantonp: it is used to map host tensors to backend tensors that are stored in a map indexed by node friendly name.

apankratovantonp: If i can extract this node from `ngraph::tensors` i can remove this structure

ilyachur: It will be better to extend ngraph::runtime::Tensor API. But in any case we cannot guarantee that HostTensor always has the name.

Do you really need to have the tensor name? Can you guarantee that tensor name exists?

apankratovantonp: This PR removes usage of tensor name in ARM Plugin

ilyachur: Yes I understood. But at the same moment [here](https://github.com/openvinotoolkit/openvino_contrib/pull/134/files#diff-c711f021618ec3aa70ee7aa1003d3d952aad1b72208f1b4c99ca94e68e1f3e3cR400) you use the friendly name to connect node with tensor. Why do you need it?

ilyachur: What if operation has several output tensors? Do these tensors have one name?

apankratovantonp: If tensor names are not used in `ngraph` reference computation code it dose not mater. We map backend tensors to elements of  `ngraph::HostTensorVector` serially one by one.

apankratovantonp: `ngraph::Tensor` name was used just to find what `ngraph::Node` produce the tensor. This PR just store the pointer to `ngraph::Node` near the tensor so we drop usage of deprecated `get_name` API

ilyachur: @apankratovantonp I really dislike such extension, because it looks like a WA. Because:

- HostTensor (if we are talking about get_name()) can be created without the name
- Tensor name can be different and it is not equal to node name
- We have name collisions if Node has several outputs

Maybe can we discuss it offline?

modules/arm_plugin/src/arm_converter/arm_converter.hpp
```diff
@@ -305,17 +310,15 @@ struct Converter {
...
                     void* host_ptr = static_cast<void*>(hostsArgument._hosts[i]->get_data_ptr());
                     void* tensor_ptr = static_cast<void*>(hostsArgument._tensors[i]->buffer());
                     if (host_ptr != tensor_ptr) {
-                        IE_SUPPRESS_DEPRECATED_START
-                        hostsArgument._hosts[i] = std::make_shared<ngraph::HostTensor>(hostsArgument._hosts[i]->get_element_type(),
-                                                               hostsArgument._hosts[i]->get_shape(),
-                                                               tensor_ptr,
-                                                               hostsArgument._hosts[i]->get_name());
-                        IE_SUPPRESS_DEPRECATED_END
+                        hostsArgument._hosts[i] = std::make_shared<ngraph::HostTensor>(
```
apankratovantonp, line 318: @ilyachur

ilyachur (approved):

modules/arm_plugin/src/arm_converter/arm_converter.hpp
ilyachur: Discussed offline.

PR #146, opened 2021-05-31
[Java] Fix failed test

likholat:


modules/java_api/tests/InferRequestTests.java
```diff
@@ -25,34 +24,34 @@ public void setUp() {
...
-    @Ignore
     @Test
     public void testGetPerformanceCounts() {
         inferRequest.Infer();
 
         Vector<String> layer_name = new Vector<>();
-        layer_name.add("19/Fused_Add_");
+        layer_name.add("19/WithoutBiases");
+        layer_name.add("20/mean/Fused_Mul_614616_const");
+        layer_name.add("20/mean/Fused_Mul_614616_const_abcd_Acdb16a_19/WithoutBiases");
```
dkurt, line 34: This is unstable check. The values are generated differently between runs and machines.

dkurt:
@likholat, simplify the test. MO is unstable but we need to test API - check that map is not empty or something like that.

dkurt (approved):

PR #149, opened 2021-06-07
[Custom layers] Custom extensions module

dkurt:


dkurt:
@ilya-lavrenov, please review. We need to put this code somewhere to enable OpenVINO SDL process for it. Contrib seems as a good alternative to keep custom layers.

PR #150, opened 2021-06-07
Added DFT and IDFT layer implementations to ArmPlugin

terfendail:


apankratovantonp (approved):

alvoron:
@terfendail I believe we forgot to update https://github.com/openvinotoolkit/openvino_contrib/blob/master/modules/arm_plugin/docs/arm_opset.md

PR #201, opened 2021-09-06
[PyTorch] Avoid origin PyTorch methods inference

dkurt:
* We still have to track tensors shapes because of `torch.split` - there is no option to define number of outputs without shapes


modules/mo_pytorch/mo_extensions/load/pytorch/hooks.py
```diff
@@ -95,78 +103,79 @@ def tensor(self):
...
-        return self._value.dim()
+        return len(self.dynamic_shape)
+    
+    @property
+    def shape(self):
+        return tuple(self.dynamic_shape)
 
     def size(self, dim=None):
-        return self._value.size(dim) if dim else self._value.size()
+        return self.dynamic_shape[dim] if dim else self.shape
```
likholat, line 116: Maybe `if dim not None`?

modules/mo_pytorch/mo_extensions/load/pytorch/detectron2/retinanet.py
```diff
@@ -6,11 +6,6 @@
 from mo.utils.error import Error
 
 def inference(model, func, anchors, pred_logits, pred_anchor_deltas, image_sizes):
```
likholat, line 8: `func` is unused parameter

dkurt: Yes, but it's used in other hooks:

```python
def forward(model, forward, batched_inputs):
    return forward([{'image': batched_inputs}])
```

likholat (approved):

PR #207, opened 2021-09-20
Replace Python assert to exception

dkurt:
related: https://github.com/openvinotoolkit/openvino/pull/6255


dkurt:
@alvoron, @likholat, please review

likholat (approved):

PR #226, opened 2021-10-23
Moved ARM plugin opset to use OV 2.0 operation RTTI

apankratovantonp:


apankratovantonp:
ping!

modules/arm_plugin/src/arm_converter/arm_converter_reduce_ops.cpp
```diff
@@ -7,11 +7,12 @@
 #include <arm_compute/runtime/NEON/functions/NEReduceMean.h>
 #include <ngraph/runtime/reference/logical_reduction.hpp>
 #include "arm_converter/arm_converter.hpp"
+#include "opset/utils.hpp"
 
 namespace ArmPlugin {
 template<typename Reduce>
 static auto ConvertReduce(const Reduce& node, const arm_compute::ReductionOperation& op, Converter* converter) {
-    auto axes = dynamic_cast<const opset::Constant&>(*(node.input_value(1).get_node())).cast_vector<int64_t>();
+    auto axes = dynamic_cast<opset::Constant&>(*(node.input_value(1).get_node())).cast_vector<std::int64_t>();
```
terfendail, line 15: Old casting approach retained

apankratovantonp: Due to some pitfalls in `ov::as_type_ptr` I have to use `dynamic_cast` here. But as it is uses reference cast it is safe enough but the error will be not so readable.

modules/arm_plugin/src/arm_converter/arm_converter.cpp
```diff
@@ -193,7 +193,7 @@ Converter::Converter(const std::shared_ptr<const ngraph::Function> function, boo
             auto hasQuantizationInfo = (itInfo != rt_info.end());
             arm_compute::TensorInfo tensorInfo;
             if (quantizedOutput && hasQuantizationInfo) {
-                auto& quantizationInfo = std::dynamic_pointer_cast<ngraph::VariantImpl<arm_compute::QuantizationInfo>>(itInfo->second)->get();
+                auto& quantizationInfo = ov::as_type_ptr<ngraph::VariantWrapper<arm_compute::QuantizationInfo>>(itInfo->second)->get();
```
terfendail, line 196: Shouldn't it be `safe_cast` as well?

modules/arm_plugin/src/transformations/quantize_fusion.cpp
```diff
@@ -95,7 +95,7 @@ ArmPlugin::pass::ConvertQuantize::ConvertQuantize() {
     register_matcher(
         std::make_shared<ngraph::pattern::Matcher>(fakeQuantize, "ConvertQuantize"),
         [](ngraph::pattern::Matcher& m) {
-            auto fakeQuantize = std::dynamic_pointer_cast<opset::FakeQuantize>(m.get_match_root());
+            auto fakeQuantize = safe_cast<opset::FakeQuantize>(m.get_match_root());
```
terfendail, line 98: Assertion that follows seems redundant

terfendail (changes requested):

modules/arm_plugin/src/transformations/quantize_fusion.cpp
apankratovantonp: done

modules/arm_plugin/src/arm_converter/arm_converter.cpp
apankratovantonp: done

modules/arm_plugin/src/arm_converter/arm_converter_gather.cpp
```diff
@@ -47,7 +47,7 @@ template <> Converter::Conversion::Ptr Converter::Convert(const ngraph::op::v1::
 }
 
 template <> Converter::Conversion::Ptr Converter::Convert(const opset::ArmGather& node) {
-    auto axes = std::dynamic_pointer_cast<opset::Constant>(node.input_value(2).get_node_shared_ptr());
+    auto axes = ov::as_type_ptr<opset::Constant>(node.input_value(2).get_node_shared_ptr());
```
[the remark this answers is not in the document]
apankratovantonp: I don't think so. Some context can add a valuable message but in other cases, such checks behave like rarely failed assertions.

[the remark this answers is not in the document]
terfendail: Here pointer value is immediately checked for `nullptr` with `IE_THROW`.
Why does this case differ from FFT implementation that use `safe_cast`?

modules/arm_plugin/src/arm_converter/arm_converter.cpp
terfendail: Looks like it's still `as_type_ptr` and unaligned with `arm_converter_conv.cpp` where `safe_cast` is used

apankratovantonp: @terfendail  Please have a look again

terfendail (approved):
Looks fine to me

PR #232, opened 2021-11-11
Using OV 2.0 RTTI

apankratovantonp:
- instead of https://github.com/openvinotoolkit/openvino_contrib/pull/226


modules/arm_plugin/src/transformations/convert_batchnorm_v0_to_v5.cpp
```diff
@@ -5,6 +5,7 @@
 #include "transformations/convert_batchnorm_v0_to_v5.hpp"
 #include "ngraph/op/batch_norm.hpp"
 #include "ngraph/rt_info.hpp"
+#include "opset/utils.hpp"
```
terfendail, line 8: Is the change necessary?

terfendail (approved):

PR #237, opened 2021-11-16
ArmPlugin LPT implementation update

terfendail:
Updated quantization implementation to support LP ResNet-50.


modules/arm_plugin/src/arm_converter/arm_converter_quantize.cpp
```diff
@@ -9,10 +9,128 @@
...
+    }
+}
 template<> Converter::Conversion::Ptr Converter::Convert(const opset::ArmQuantize& node) {
-    return MakeConversion<arm_compute::NEQuantizationLayer>(node.input(0), node.output(0));
+    auto qInfoIt = node.get_rt_info().find("QuantizationInfo");
+    if (qInfoIt == node.get_rt_info().end()) {
+        IE_THROW() << "No quantization info available for ArmQuantize";
+    }
+    auto qInfo = safe_cast<ngraph::VariantWrapper<arm_compute::QuantizationInfo>>(qInfoIt->second)->get();
+    if (node.get_input_element_type(0).is_real()) {
```
apankratovantonp, line 73: Move this check into transformation

modules/arm_plugin/src/arm_converter/arm_converter_quantize.cpp
```diff
@@ -9,10 +9,128 @@
...
+                                        node.input(0),
+                                        node.output(0),
+                                        node.get_input_shape(0),
+                                        qInfo.scale()[0],
+                                        qInfo.offset()[0]);
+        };
+        return CallSwitch(
+            AP_WRAP(make, quantize),
+            node.get_input_element_type(0), floatTypes,
+            node.get_output_element_type(0), intTypes);
```
apankratovantonp, line 85: Do we need quantization into all integer types?

terfendail: fixed

modules/arm_plugin/src/arm_converter/arm_converter_quantize.cpp
terfendail: That's not a call to reference implementation. It's a local implementation that will require definition of separate operation to be controlled via transformations. I think it doesn't make sense to define two separate ArmQuantize operations used for non-intersecting data type sets just to make possible selection of a more proper implementation of the same operation in runtime.
By the way I'm going to investigate whether it would be possible to avoid use of scalar implementation at all.

apankratovantonp (approved):

modules/arm_plugin/src/arm_converter/arm_converter_conv.cpp
```diff
@@ -59,14 +254,22 @@ template<> Converter::Conversion::Ptr Converter::Convert(const opset::ArmGroupCo
         ngraphWeightsShape[3],
         ngraphWeightsShape[4]
     }));
+
+    auto wInfoIt = node.get_rt_info().find("WeightsPrescaleInfo");
+    arm_compute::QuantizationInfo* wInfo = wInfoIt == node.get_rt_info().end() ? nullptr :
+                                           &(safe_cast<ngraph::VariantWrapper<arm_compute::QuantizationInfo>>(wInfoIt->second)->get());
```
apankratovantonp, line 260: Use arm_compute::QuantizationInfo directly without `ngraph::VariantWrapper`

modules/arm_plugin/src/arm_converter/arm_converter_conv.cpp
```diff
@@ -59,14 +254,22 @@ template<> Converter::Conversion::Ptr Converter::Convert(const opset::ArmGroupCo
         ngraphWeightsShape[3],
         ngraphWeightsShape[4]
     }));
+
+    auto wInfoIt = node.get_rt_info().find("WeightsPrescaleInfo");
```
apankratovantonp, line 258: BTW: If you need to store input/output specific runtime info you can do it in input/output as they also has `rt_info`

modules/arm_plugin/src/transformations/quantize_fusion.cpp
```diff
@@ -192,14 +198,38 @@ ArmPlugin::pass::NodeQuantizeFusion::NodeQuantizeFusion() {
...
+
             std::vector<ngraph::Output<ngraph::Node>> newInputs;
             Types inputTypes;
             for (auto&& input : node->inputs()) {
                 inputTypes.emplace_back(realType);
                 newInputs.emplace_back(
                     ngraph::op::TemporaryReplaceOutputType{input.get_source_output(), realType}.get());
             }
+            std::int32_t qiOffset = 0;
+            if (!std::all_of(std::begin(quantizationInfo.second), std::end(quantizationInfo.second), [&] (auto offset) {
```
apankratovantonp, line 214: Please use named bool instead if placing a long call into `if()`

modules/arm_plugin/src/arm_converter/arm_converter_conv.cpp
terfendail: fixed

modules/arm_plugin/src/arm_converter/arm_converter_conv.cpp
terfendail: Even though for ACL quantization info relates to tensors from the ngraph point of view these properties are internal to operation, so I prefer to retain them as part of the node itself.

modules/arm_plugin/src/transformations/quantize_fusion.cpp
terfendail: fixed

modules/arm_plugin/src/arm_converter/arm_converter_mat_mul.cpp
```diff
@@ -3,21 +3,207 @@
...
 
 
+#include <src/cpu/kernels/CpuConvertQuantizedSignednessKernel.h>
+#include <arm_compute/runtime/NEON/NEScheduler.h>
 #include <arm_compute/runtime/NEON/functions/NEFullyConnectedLayer.h>
 #include "arm_converter/arm_converter.hpp"
 
 namespace ArmPlugin {
 enum InputArg {Features, Weights, Bias};
+struct NEFullyConnectedLayerQI final: public arm_compute::IFunction {
```
apankratovantonp, line 13: Can we move this implementation to a separate subfolder? 
Can we generalize some code inside them? Makes some base `LayerQI` class with some common parts of code?
It will be good if these parts were moved into separate functions with clear names.

terfendail: I think these wrappers shouldn't be separated from the conversions they are used in. At least such a structure was used for all other wrappers introduced to support reference based operations. But I definitely agree with simplification of wrappers via generalized class.
However I suppose that refactoring could take some time so I prefer to do it as a part of separate PR to improve LPT support ASAP and avoid possible merge conflicts due to the massiveness of this PR code change.

PR #238, opened 2021-11-16
[ARM Plugin] added adaptive_pooling reference implementation

rayonnant14:


modules/arm_plugin/src/arm_converter/arm_converter_adaptive_pooling.cpp
```diff
@@ -0,0 +1,38 @@
...
+                                node.input(0),
+                                node.output(0),
+                                node.output(1),
+                                node.get_input_shape(0),
+                                node.get_output_shape(0));
+    };
+    return CallSwitch(
+        AP_WRAP(make, ngraph::runtime::reference::adaptive_max_pool),
+        node.input(0), floatTypes,
+        node.input(1), indexTypes);
```
terfendail, line 35: Shouldn't it be `node.output(1)`?

terfendail (changes requested):

modules/arm_plugin/src/arm_converter/arm_converter_adaptive_pooling.cpp
rayonnant14: fixed, thank you!

terfendail (approved):

PR #259, opened 2021-12-08
[ARM Plugin] added color_convert_nv12 reference support

rayonnant14:


modules/arm_plugin/src/arm_converter/arm_converter_convert_color_nv12.cpp
```diff
@@ -0,0 +1,72 @@
...
+    auto make = [&] (auto refFunction) {
+        return this->MakeConversion(refFunction,
+                                    node.input(0),
+                                    node.get_input_shape(0),
+                                    single_plane ? node.input(0) : node.input(1),
+                                    node.output(0),
+                                    ov::op::util::ConvertColorNV12Base::ColorConversion::NV12_TO_BGR,
+                                    single_plane);
+        };
+    constexpr static auto types = std::tuple<std::uint8_t, ngraph::float16, float>{};
```
terfendail, line 49: IMO it would be better to move `std::tuple<...>{}` directly to `CallSwitch`
Anyway `types` if retained shouldn't be defined as `static` while `constexpr` is just unnecessary here.

terfendail (approved):

PR #264, opened 2021-12-14
Workaround MO bug

dkurt:
ticket 73794


dkurt:
wait for https://github.com/openvinotoolkit/openvino/pull/9206

PR #270, opened 2021-12-18
[ARM-plugin] Fix clang RTTI problem on Android

allnes:


modules/arm_plugin/src/transformations/arm_optimizations.cpp
```diff
@@ -97,6 +97,13 @@
 
 #include "arm_optimizations.hpp"
 
+#if(__ANDROID__)
+#define REGISTER_PASS manager.register_pass<ov::pass::GraphRewrite>()->add_matcher
```
ilyachur, line 101: @GlebKazantaev it looks like a WA

apankratovantonp:
Needed More detailed explanation

modules/arm_plugin/src/transformations/arm_optimizations.cpp
```diff
@@ -242,84 +249,83 @@ bool ArmPlugin::pass::ArmOptimizations::run_on_function(std::shared_ptr<ngraph::
...
         manager.register_pass<ngraph::pass::ConvertBroadcastToTiles>();
-        manager.register_pass<pass::ConvertEltwise>();
+        REGISTER_PASS<pass::ConvertEltwise>();
         manager.register_pass<ngraph::pass::ConstantFolding>();
-        manager.register_pass<pass::ConvertTile>();
-        manager.register_pass<pass::ConvertSplit>();
-        manager.register_pass<pass::ConvertConcat>();
+        REGISTER_PASS<pass::ConvertTile>();
+        REGISTER_PASS<pass::ConvertSplit>();
+        REGISTER_PASS<pass::ConvertConcat>();
```
terfendail, line 307: Lines 308 and 309 looks local as well

modules/arm_plugin/src/transformations/arm_optimizations.cpp
```diff
@@ -153,7 +160,7 @@ bool ArmPlugin::pass::ArmOptimizations::run_on_function(std::shared_ptr<ngraph::
     manager.register_pass<ngraph::pass::RemoveFilteringBoxesBySize>(); // Resolves dynamism (replaces NonZero), CF needed
     manager.register_pass<ngraph::pass::ConstantFolding>();
     manager.register_pass<ngraph::pass::NopElimination>(); // may introduce fake dynamism
-    manager.register_pass<pass::ReplacePowerByMul>();
+    REGISTER_PASS<pass::ReplacePowerByMul>();
```
terfendail, line 163: Line 159 looks local as well

allnes:
@apankratovantonp 
Our manual arm quantization transformations couldn't run on arm android platform (for example 
[error.txt](https://github.com/openvinotoolkit/openvino_contrib/files/7744976/error.txt)), this workaround `manager.register_pass<ov::pass::GraphRewrite>()->add_matcher`  fix it. But this fix break inference on RPI with same errors. That why we create macros
```
#if(__ANDROID__)
#define REGISTER_PASS manager.register_pass<ov::pass::GraphRewrite>()->add_matcher
#else
#define REGISTER_PASS manager.register_pass
#endif
```

alvoron:
@apankratovantonp I assume we don't consider merging this PR. It's being used as an example how we avoid dynamic_pointer_castn issue in ARM plugin to discuss the issue with nGraph guys.

allnes:
@terfendail @GlebKazantaev please review

modules/arm_plugin/src/transformations/arm_optimizations.cpp
```diff
@@ -136,73 +136,74 @@ void ArmPlugin::pass::ArmOptimizations::Dump(const std::shared_ptr<ngraph::Funct
...
+                std::vector<ngraph::element::Type>{ ngraph::element::i8, ngraph::element::u8 });
+        }
 
+        // This pass must be called first in pipeline
+        manager.register_pass<ngraph::pass::InitNodeInfo>();
+        manager.register_pass<pass::StoreResultName>();
+        manager.register_pass<ngraph::pass::RemoveFilteringBoxesBySize>(); // Resolves dynamism (replaces NonZero), CF needed
+        manager.register_pass<ngraph::pass::ConstantFolding>();
+        manager.register_pass<ngraph::pass::NopElimination>(); // may introduce fake dynamism
+        manager.register_pass<pass::ReplacePowerByMul>();
```
terfendail, line 157: `ReplacePowerByMul` is a local `MatcherPass` also

modules/arm_plugin/src/transformations/arm_optimizations.cpp
```diff
@@ -136,73 +136,74 @@ void ArmPlugin::pass::ArmOptimizations::Dump(const std::shared_ptr<ngraph::Funct
...
+        // LinOpSequenceFusion must be executed after all decompositions
+        manager.register_pass<ngraph::pass::LinOpSequenceFusion>();
+        manager.register_pass<ngraph::pass::RNNCellDecomposition>();
+        manager.register_pass<ngraph::pass::LSTMCellDecomposition>();
+        manager.register_pass<ngraph::pass::GRUCellDecomposition>();
+        manager.register_pass<ngraph::pass::ConvertGELU>();
+        manager.register_pass<ngraph::pass::ConstantFolding>();
+        manager.register_pass<pass::ConvertConv1D>();
+        manager.register_pass<pass::ConvertGroupConv1D>();
+        manager.register_pass<pass::ConvertGroupConvolution>();
```
terfendail, line 172: Local `MatcherPass`-es also

modules/arm_plugin/src/transformations/arm_optimizations.cpp
```diff
@@ -242,7 +243,7 @@ bool ArmPlugin::pass::ArmOptimizations::run_on_function(std::shared_ptr<ngraph::
 
     {
         Dump(f, "before_arm_specific_transformations");
-        ngraph::pass::Manager manager;
+        ov::pass::Manager manager;
```
terfendail, line 246: Lines 248-251, 261-262,265 and probably some more are local `MatcherPass`-es also

modules/arm_plugin/src/transformations/arm_optimizations.cpp
```diff
@@ -309,19 +310,34 @@ bool ArmPlugin::pass::ArmOptimizations::run_on_function(std::shared_ptr<ngraph::
...
-        manager.run_passes(f);
+        {
+            ov::pass::Manager manager;
+            auto pass = manager.register_pass<ov::pass::GraphRewrite>();
+            pass->add_matcher<pass::ConvolutionQuantizeFusion>();
+            pass->add_matcher<pass::MeanQuantizeFusion>();
+            manager.run_passes(f);
+        }
+        {
+            ov::pass::Manager manager;
```
terfendail, line 323: I think it would be better to use single `Manager` for whole arm quantization processing and execute `run_passes` once

terfendail (changes requested):
Even though the issue is easily reproduceable for quantized nets only the root cause is generic. So I prefer to wrap with `GraphRewrite` all registrations for local `MatcherPass` instead of just that we combine for simultaneous execution.
I don't think the overhead will be high while we ensure all transformations are working as expected

apankratovantonp:
Seems that we do not export RTTI for this classes as we do not export these classes from library too.
We can export our local pass symbols from library: @ilya-lavrenov @ilyachur  could you advise us how we should do it:
- Should we produce our local API macro or we can use `OPENVINO_API` macro. How should we implement it. 
- Or should we use OPENVINO_RTTI macro? 
- @allnes  we should apply the API macro to all passes objects
- @allnes as we dose not override any virtual functions maybe we need to override at least the destructor and place its default implementation into cpp file for each pass.

@ilya-lavrenov @GlebKazantaev we need to add this case into documentation

apankratovantonp (approved):

allnes:
@terfendail @apankratovantonp @alvoron please review

terfendail (approved):

PR #277, opened 2022-01-10
Remove USRNet from the list of supported networks

alvoron:
As a result of this PR: https://github.com/openvinotoolkit/openvino_contrib/pull/261


dkurt:
Do we really need to remove the model name from supported list?

alvoron:
@dkurt I don't think it's required from a security perspective. I just thought to make README aligned with tests.
If you prefer to keep it in README I think we may close this PR.

dkurt:
Yes, I just wanted to keep a reference to USRNet if there is no legal/security issues here.

PR #286, opened 2022-01-20
[ARM Plugin] avgPool support for 5D tensors

rayonnant14:


modules/arm_plugin/src/arm_converter/arm_converter_pool.cpp
```diff
@@ -191,16 +192,34 @@ struct NEPoolingLayerQI final: public arm_compute::IFunction {
...
+        pool_info.exclude_padding = node.get_exclude_pad();
+        auto iInfoIt = node.get_rt_info().find("InputPrescaleInfo");
+        const arm_compute::QuantizationInfo* iInfo = iInfoIt == node.get_rt_info().end() ? nullptr :
+                                                &(iInfoIt->second.as<arm_compute::QuantizationInfo>());
+        auto qInfoIt = node.get_rt_info().find("QuantizationInfo");
+        const arm_compute::QuantizationInfo* qInfo = qInfoIt == node.get_rt_info().end() ? nullptr :
+                                                &(qInfoIt->second.as<arm_compute::QuantizationInfo>());
+        return MakeConversion<NEPoolingLayerQI>(node.input(0), node.output(0), pool_info, iInfo, qInfo);
+    } else {
+        auto make = [&] (auto refFunction) {
```
terfendail, line 208: Could you please add here assertions that neither `InputPrescaleInfo` nor `QuantizationInfo` are provided for the node.

Also it would be necessary to check input shape at
https://github.com/openvinotoolkit/openvino_contrib/blob/cdaa9274680fe15cede13bb3d76209b663d13211/modules/arm_plugin/src/transformations/quantize_fusion.cpp#L259
and
https://github.com/openvinotoolkit/openvino_contrib/blob/cdaa9274680fe15cede13bb3d76209b663d13211/modules/arm_plugin/src/transformations/quantize_fusion.cpp#L397
to avoid conversion of unsuitable operations and creation of quantization information for the node

terfendail (changes requested):

modules/arm_plugin/src/transformations/quantize_fusion.cpp
```diff
@@ -387,6 +390,9 @@ ArmPlugin::pass::DequantizeInputFusion::DequantizeInputFusion() {
             }
 
             auto node = output->output(0).get_target_inputs().begin()->get_node()->shared_from_this();
+            if (ngraph::is_type<opset::AvgPool>(node) && node->get_input_shape(0).size() == 5) {
```
terfendail, line 393: It would be better to move this check inside the first if branch.
Network could contain unquantized AvgPool layer that at the moment will be processed by the second branch and prepended with Dequantize layer instead of scale and/or shift layers

terfendail (changes requested):

modules/arm_plugin/src/transformations/quantize_fusion.cpp
```diff
@@ -349,6 +352,10 @@ ArmPlugin::pass::DequantizeInputFusion::DequantizeInputFusion() {
             if (output->output(0).get_target_inputs().size() != 1 || !realType.is_real()) {
                 return false;
             }
+            auto node = output->output(0).get_target_inputs().begin()->get_node()->shared_from_this();
+            if (ngraph::is_type<opset::AvgPool>(node) && node->get_input_shape(0).size() == 5) {
```
terfendail, line 356: It would be better to move this check inside `if` at line 390/395.
This check ensure QInfo will be configured for AvgPool with INT8 output, while AvgPool with F32 output will be processed by other branch.
So for the sequence U8->Mul-(F32)->AvgPool->F32 transformation will yield
U8->Dequantize-(f32)->AvgPool->F32
instead of U8->Dequantize-(F32)->Mul-(F32)->AvgPool->F32 prepared by `DequantizeOnInputs` transformation that follow

PR #295, opened 2022-01-26
[ARM Plugin] Ubuntu 20 support

alvoron:
This docker image is the same as for 18 ubuntu except for sources.list content.
I tested it on my 20.04, it works fine, only python 3.7 needs to be installed manually (20.04 comes with 3.8).

### DIFF
--- a/modules/arm_plugin/Dockerfile.RPi64_focal
+++ b/modules/arm_plugin/Dockerfile.RPi64_focal
@@ -1,0 +1,108 @@
+#This Dockerfile is for x86 and should be used for OpenVINO ARM plugin cross-compilation
+#https://github.com/openvinotoolkit/openvino_contrib/tree/master/modules/arm_plugin#how-to-build
+
+#Some pieces of Dokerfile are taken from OpenVINO CI image:
+#https://github.com/intel-innersource/frameworks.ai.openvino.qa.dockerfiles/blob/master/dldt_build/debian_9_arm/Dockerfile
+
+FROM ubuntu:20.04
+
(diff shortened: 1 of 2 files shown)


allnes (approved):

modules/arm_plugin/Dockerfile.RPi64_focal
```diff
@@ -0,0 +1,108 @@
...
+        --host=aarch64-linux-gnu \
+        --build=x86_64-linux-gnu \
+        --disable-ipv6 \
+        --enable-shared \
+        --prefix=/opt/python3.7_arm \
+        ac_cv_file__dev_ptmx=no \
+        ac_cv_file__dev_ptc=no && \
+    make -j4 && make altinstall
+
+RUN wget https://www.cmake.org/files/v3.13/cmake-3.13.4.tar.gz && \
```
terfendail, line 82: Ubuntu Focal has cmake 3.16.3 so building from source is unnecessary

terfendail:
Also I think we could try to stick to python 3.8 instead of 3.7 building. However it's not urgent

modules/arm_plugin/Dockerfile.RPi64_focal
alvoron: removed cmake compilation

alvoron:
> Also I think we could try to stick to python 3.8 instead of 3.7 building. However it's not urgent

I'll fix it in the next PR. IMO we shouldn't hardcode python-related paths, so the script may pick up any python version correctly.

PR #296, opened 2022-01-26
[ARM Plugin] Fix klocwork issue

alvoron:


modules/arm_plugin/src/transformations/finalize_trailing_nodes.cpp
```diff
@@ -19,9 +19,11 @@ bool ArmPlugin::pass::FinalizeTrailingNodes::run_on_function(std::shared_ptr<ngr
             auto inputs = out.get_target_inputs();
             if (inputs.empty() && !std::dynamic_pointer_cast<opset::Result>(node)) {
                 auto result = std::make_shared<opset::Result>(out);
-                result->set_friendly_name(ngraph::op::util::get_ie_output_name(out));
-                f->add_results({result});
-                is_modified = true;
+                if (result) {
```
alalek, line 22: > `std::make_shared`

This call can't return empty pointer (it throws `bad_alloc` exception or another exception from constructor).
So this check is redundant.

It is a Klocwork FA, it messes with understanding of STL library.
/cc @asomsiko Is there robust solution for this?

alvoron: Thank you for letting me know. I'll close the PR and mark it in Klocwork as false positive.

PR #300, opened 2022-02-01
Fix the skip pattern for hanging ARM test

akladiev:
Looks like the issue described in 72140 still reproduces on the same test case, but without '_' postfix


alvoron (approved):

alvoron:
@akladiev could you please pull the latest changes from master? It contains the fix for Azure CI - right now it blocks us from the merge.

PR #307, opened 2022-02-07
Migrated Arm plugin to ov::Model

terfendail:


apankratovantonp (approved):
Grate that variables are renamed too!

PR #311, opened 2022-02-11
[ARM Plugin] maxpool-v8 transformation

rayonnant14:


modules/arm_plugin/src/arm_converter/arm_converter_pool.cpp
```diff
@@ -57,24 +59,35 @@ template<> Converter::Conversion::Ptr Converter::Convert(const opset::MaxPool& n
...
-    };
-    return CallSwitch(
-        AP_WRAP(make, ngraph::runtime::reference::max_pool),
-        node.input(0), allTypes,
-        node.output(1), indexTypes);
+    if ((node.get_input_shape(0).size() == 4) &&
+       (node.get_output_element_type(1) == ngraph::element::u32) &&
+       (node.get_kernel() == ngraph::Shape{2, 2}) &&
+       (node.get_dilations() == ngraph::Strides{1, 1}) &&
+       (node.get_axis() == 0)) {
```
terfendail, line 66: I think such a strong restriction is too tough here. We could use ACL whenever there are no consumers for the second output

modules/arm_plugin/src/transformations/convert_maxpool_v8.cpp
```diff
@@ -0,0 +1,57 @@
...
+                                                                     maxpool->get_axis());
+        new_ops.emplace_back(new_maxpool);
+
+        ngraph::Output<ngraph::Node> output_0 = new_maxpool->output(0);
+        ngraph::Output<ngraph::Node> output_1_convert = new_maxpool->output(1);
+
+        output_1_convert = std::make_shared<opset::Convert>(output_1_convert,
+                                                            maxpool->get_output_element_type(1));
+        output_1_convert.get_node_shared_ptr()->set_friendly_name(maxpool->get_friendly_name() + "/convert.1");
+        new_ops.emplace_back(output_1_convert.get_node_shared_ptr());
```
terfendail, line 50: `new_ops` vector isn't actually used. Could it be removed?

modules/arm_plugin/src/arm_converter/arm_converter_pool.cpp
terfendail: This change could be postponed to a separate PR as discussed

modules/arm_plugin/src/transformations/convert_maxpool_v8.cpp
```diff
@@ -0,0 +1,54 @@
...
+                                                                     maxpool->get_strides(),
+                                                                     maxpool->get_dilations(),
+                                                                     maxpool->get_pads_begin(),
+                                                                     maxpool->get_pads_end(),
+                                                                     maxpool->get_kernel(),
+                                                                     maxpool->get_rounding_type(),
+                                                                     maxpool->get_auto_pad(),
+                                                                     ngraph::element::u32,
+                                                                     maxpool->get_axis());
+
```
terfendail, line 41: Redundant whitespace here

alvoron:
@terfendail it seems all code style issues have been fixed. Do you have any other comments or we could proceed with merge?

terfendail (approved):

terfendail:
There are a few failed builds but all required are passed and all arm builds are passed as well. Probably failed builds relate to buildfarm issues so lets merge the PR

PR #346, opened 2022-03-16
Test contrib smart ci

akladiev:


alvoron:
@akladiev is it still needed? may we close it?

akladiev:
> @akladiev is it still needed? may we close it?

Yes, will close it for now

PR #349, opened 2022-03-22
[ARM-plugin]  support for Interpolate node with 3 inputs

rayonnant14:


modules/arm_plugin/src/arm_converter/arm_converter_interpolate.cpp
```diff
@@ -78,24 +75,37 @@ void wrap_interpolate(const T* input_data,
                                                 attrs);
 }
 
+void get_axes_vec(const opset::Interpolate& node, std::vector<int64_t>& axes_vec) {
+    if (node.get_input_size() == 4) {
+        auto axes = safe_cast<opset::Constant>(node.input_value(3).get_node_shared_ptr());
+        axes_vec = axes->cast_vector<int64_t>();
+    } else {
+        ngraph::PartialShape input_shape = ngraph::PartialShape(node.get_input_partial_shape(0));
```
terfendail, line 83: Could we use just `auto input_shape = node.get_input_partial_shape(0);` here?

modules/arm_plugin/src/transformations/convert_interpolate_arm.cpp
```diff
@@ -98,6 +98,21 @@ ArmPlugin::pass::ConvertInterpolate::ConvertInterpolate() {
             return false;
         }
 
+        if (interp->get_input_size() == 4) {
```
terfendail, line 101: Why do we need to ensure the default axes order? As far as I could understand axes order is important only for input(1) and input(2) which aren't used anyway by ArmInterpolate since it rely on correct output(0) shape inference provided by IE core.

modules/arm_plugin/src/transformations/convert_interpolate_arm.cpp
```diff
@@ -109,7 +124,6 @@ ArmPlugin::pass::ConvertInterpolate::ConvertInterpolate() {
         auto arm_interp = std::make_shared<opset::ArmInterpolate>(interp->input_value(0),
                                                                   interp->input_value(1),
                                                                   interp->input_value(2),
-                                                                  interp->input_value(3),
```
terfendail, line 112: Looks like input_value(1) and input_value(2) could be removed as well. Neither of them are used inside ArmInterpolate conversion

modules/arm_plugin/src/arm_converter/arm_converter_interpolate.cpp
```diff
@@ -78,24 +75,37 @@ void wrap_interpolate(const T* input_data,
                                                 attrs);
 }
 
+void get_axes_vec(const opset::Interpolate& node, std::vector<int64_t>& axes_vec) {
+    if (node.get_input_size() == 4) {
+        auto axes = safe_cast<opset::Constant>(node.input_value(3).get_node_shared_ptr());
```
terfendail, line 80: I suppose it's a really rare case but anyway it's possible that axis input wouldn't be a constant.
I prefer to stick to previous processing scheme for the case of 4 inputs to retain non-const axis support

modules/arm_plugin/src/transformations/convert_interpolate_arm.cpp
rayonnant14: Right, I removed this check

modules/arm_plugin/src/arm_converter/arm_converter_interpolate.cpp
```diff
@@ -56,46 +65,67 @@ void wrap_interpolate(const T* input_data,
...
+                                               attrs);
 }
 
 template<> Converter::Conversion::Ptr Converter::Convert(const opset::Interpolate& node) {
     auto make = [&] (auto refFunction) {
+    if (node.get_input_size() == 4) {
         return this->MakeConversion(refFunction,
                                     node.input(0),
                                     node.get_input_shape(0),
+                                    node.get_input_partial_shape(0),
```
terfendail, line 102: `get_input_shape()` infers `Shape` from partial shape if it's static and throw exception otherwise. Usually it's safer to call for partial shape, but we anyway use static shape already so call for partial shape is redundant

modules/arm_plugin/src/transformations/convert_interpolate_arm.cpp
terfendail: Could it be reduced to `arm_interp = std::make_shared<opset::ArmInterpolate>(interp->input_value(0), attrs);`?

alvoron (approved):

alvoron:
Jenkins CI is passed, however its status hasn't been reported to github.

Issue #350, opened 2022-03-23
Fix failed Java tests


dkurt:
https://dev.azure.com/openvinoci/dldt/_build/results?buildId=307827&view=logs&j=0609cd82-f9f7-5e70-371c-89fca78031ab&t=73ab56eb-fe05-59b6-aeb0-ce6e8e3090a3
```
tests.CoreTests > testReadModelIncorrectBinPath STANDARD_OUT
    testReadModelIncorrectBinPath(tests.CoreTests)- [CPU] - FAILED

tests.CoreTests > testReadModelIncorrectBinPath FAILED
    java.lang.AssertionError
        at org.junit.Assert.fail(Assert.java:87)
        at org.junit.Assert.assertTrue(Assert.java:42)
        at org.junit.Assert.assertTrue(Assert.java:53)
        at tests.CoreTests.testReadModelIncorrectBinPath(CoreTests.java:34)
```
```
tests.compatibility.IECoreTests > testReadNetworkIncorrectBinPath STANDARD_OUT
    testReadNetworkIncorrectBinPath(tests.compatibility.IECoreTests)- [CPU] - FAILED

tests.compatibility.IECoreTests > testReadNetworkIncorrectBinPath FAILED
    java.lang.AssertionError
        at org.junit.Assert.fail(Assert.java:87)
        at org.junit.Assert.assertTrue(Assert.java:42)
        at org.junit.Assert.assertTrue(Assert.java:53)
        at tests.compatibility.IECoreTests.testReadNetworkIncorrectBinPath(IECoreTests.java:45)
```

PR #351, opened 2022-03-23
[JAVA] Fixed exception message in tests

likholat:
https://github.com/openvinotoolkit/openvino_contrib/issues/350#event-6289159371

### DIFF
--- a/modules/java_api/tests/CoreTests.java
+++ b/modules/java_api/tests/CoreTests.java
@@ -31,7 +31,7 @@ public class CoreTests extends OVTest {
         } catch (Exception e) {
             exceptionMessage = e.getMessage();
         }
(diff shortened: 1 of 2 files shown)


modules/java_api/tests/CoreTests.java
```diff
@@ -31,7 +31,7 @@ public void testReadModelIncorrectBinPath() {
         } catch (Exception e) {
             exceptionMessage = e.getMessage();
         }
-        assertTrue(exceptionMessage.contains("Weights file model.bin cannot be opened!"));
+        assertTrue(exceptionMessage.contains("Can not open file model.bin for mapping."));
```
dkurt, line 34: Can we replace to more flexible check that `exceptionMessage` is not empty?

dkurt (approved):

PR #352, opened 2022-03-23
Reduce MaxPoolV8 to MaxPoolV1 where possible

terfendail:
MaxPoolV8 should be inferred via ACL(V1 compatible) implementation where possible to avoid performance degradation on recent IRs that usually contain V8 even if V1 is suitable

### DIFF
--- a/modules/arm_plugin/src/arm_converter/arm_converter_pool.cpp
+++ b/modules/arm_plugin/src/arm_converter/arm_converter_pool.cpp
@@ -7,7 +7,6 @@
 #include <arm_compute/runtime/NEON/functions/NEPoolingLayer.h>
 #include <ngraph/runtime/reference/max_pool.hpp>
 #include <ngraph/runtime/reference/avg_pool.hpp>
-#include <transformations/utils/utils.hpp>
 #include "arm_converter/arm_converter.hpp"
 
 
--- a/modules/arm_plugin/src/arm_converter/arm_converter_pool.cpp
+++ b/modules/arm_plugin/src/arm_converter/arm_converter_pool.cpp
@@ -70,16 +70,6 @@ template<> Converter::Conversion::Ptr Converter::Convert(const ngraph::op::v8::M
         FillLayerInfo(node, pool_info);
         pool_info.pool_type = arm_compute::PoolingType::MAX;
         return MakeConversion<arm_compute::NEPoolingLayer>(node.input(0), node.output(0), pool_info, node.output(1));
-    } else if ((node.get_input_shape(0).size() == 4) &&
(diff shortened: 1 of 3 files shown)


modules/arm_plugin/src/arm_converter/arm_converter_pool.cpp
```diff
@@ -69,6 +70,16 @@ template<> Converter::Conversion::Ptr Converter::Convert(const ngraph::op::v8::M
         FillLayerInfo(node, pool_info);
         pool_info.pool_type = arm_compute::PoolingType::MAX;
         return MakeConversion<arm_compute::NEPoolingLayer>(node.input(0), node.output(0), pool_info, node.output(1));
+    } else if ((node.get_input_shape(0).size() == 4) &&
```
rayonnant14, line 73: In this case also possible to use transformation for `ngraph::op::v8::MaxPool` like `ArmPlugin::pass::ConvertMaxPool1D::ConvertMaxPool1D()` to process 3D input shapes too

terfendail: I've replaced special processing with V8->V1 transformation. Now all possible cases should be handled if they are supported by V1

modules/arm_plugin/src/transformations/arm_optimizations.cpp
```diff
@@ -277,9 +277,9 @@ bool ArmPlugin::pass::ArmOptimizations::run_on_function(std::shared_ptr<ov::Mode
         manager.register_pass<ov::pass::GraphRewrite>()->add_matcher<pass::ConvertInterpolate>();
         manager.register_pass<ov::pass::GraphRewrite>()->add_matcher<pass::ConvertMVN>();
         manager.register_pass<ov::pass::GraphRewrite>()->add_matcher<pass::ConvertReorgYolo>();
+        manager.register_pass<ov::pass::GraphRewrite>()->add_matcher<pass::ConvertMaxPoolV8>();
```
rayonnant14, line 280: This changes looks similar to this [transformation](https://github.com/openvinotoolkit/openvino/blob/master/src/common/transformations/src/transformations/op_conversions/convert_maxpool_downgrade.cpp). Maybe reuse it?
```cpp
manager.register_pass<ov::pass::GraphRewrite>()->add_matcher<ngraph::pass::ConvertMaxPool8ToMaxPool1>();
```

terfendail: Thanks. I missed the standard one. That's exactly what is required

rayonnant14 (approved):

PR #383, opened 2022-06-27
[ovms_ai_extension] Adding in new OVMS AI Extension module

DrewBearly:
Adding in the OpenVINO Model Server AI Extension to openvino_contrib


dkalinowski:
Why do we want to have OVMS source code in openvino_contrib?
There is separate github repository for OVMS:
https://github.com/openvinotoolkit/model_server/tree/main

What's more, I can see the code you pushed is quite old, probably from `master` branch. OVMS uses branch `main` as its latest branch. Master branch was frozen in 2019 when we switched from python to C++.

dkalinowski:
> > Why do we want to have OVMS source code in openvino_contrib? There is separate github repository for OVMS: https://github.com/openvinotoolkit/model_server/tree/main
> > What's more, I can see the code you pushed is quite old, probably from `master` branch. OVMS uses branch `main` as its latest branch. Master branch was frozen in 2019 when we switched from python to C++.
> 
> Hi. This is the new version of OVMS extension - this API was not developed longer by OVMS team and we decided to continue maintaining it. As we dont have a separate repo, this location is agreed with OV Toolkit Product Management.

I see. My only concern about this PR is that it might contain license files for multiple components which I think are not used by final image (for example gcc). Other than that looks good to me.

ryanloney:
@dtrawins @mzegla could one of you take a quick look before this is merged. The Azure extension module is being moved here.

helena-intel:
@DrewBearly can you change the description in the main README to make it clearer what this extension does - what it enables that is not in regular OVMS? It would also be good to clarify in both main and module README that this is for 2021.4.

ryanloney:
Thanks @helena-intel. This part of the README does not make any sense. We should be clear this is an Extension for some Azure product. 
image

kerdanow:
> > > Why do we want to have OVMS source code in openvino_contrib? There is separate github repository for OVMS: https://github.com/openvinotoolkit/model_server/tree/main
> > > What's more, I can see the code you pushed is quite old, probably from `master` branch. OVMS uses branch `main` as its latest branch. Master branch was frozen in 2019 when we switched from python to C++.
> > 
> > 
> > Hi. This is the new version of OVMS extension - this API was not developed longer by OVMS team and we decided to continue maintaining it. As we dont have a separate repo, this location is agreed with OV Toolkit Product Management.
> 
> I see. My only concern about this PR is that it might contain license files for multiple components which I think are not used by final image (for example gcc). Other than that looks good to me.

If any component which we are not using in the final image, we can remove it from the license file. With that said, it seems like gcc is a linked dependency of the “build-essential” package, which is a part of Docker file. So adding that license is an obligation we need to fulfill

modules/ovms_ai_extension/Dockerfile
```diff
@@ -0,0 +1,37 @@
+# Copyright (C) 2021 Intel Corporation
+
+FROM openvino/model_server:2021.4.1-gpu
+USER root 
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    build-essential \
```
mzegla, line 6: Is this package used? It seems redundant.

modules/ovms_ai_extension/Dockerfile
```diff
@@ -0,0 +1,37 @@
+# Copyright (C) 2021 Intel Corporation
+
+FROM openvino/model_server:2021.4.1-gpu
```
mzegla, line 3: Have you considered using latest OVMS image?

modules/ovms_ai_extension/src/ams_wrapper/requirements.txt
```diff
@@ -0,0 +1,6 @@
+tensorflow-serving-api==2.7.0
```
mzegla, line 1: There's a lightweight Python client for OVMS available. See:
- https://pypi.org/project/ovmsclient/
- https://community.intel.com/t5/Blogs/Tech-Innovation/Artificial-Intelligence-AI/Simplified-Deployments-with-OpenVINO-Model-Server-and-TensorFlow/post/1353218

Switching to `ovmsclient` will let you significantly reduce overall Docker image size. But this gain is possible only, if you drop tensorflow-serving-api and all it's dependencies (including tensorflow itself).

DrewBearly:
Apologies for the delay on this - bit of vacation got right in the way. Expect requested documentation updates to be taken care of today and tomorrow, with new info about exactly what, why, and how Azure is used with this and resolving other open items. Appreciate the patience on this!

modules/ovms_ai_extension/Dockerfile
DrewBearly: We pegged against a specific version for testing/validation purposes. We may update in the future to just use the latest image, but for now we're keeping this.

modules/ovms_ai_extension/Dockerfile.lgpl
```diff
@@ -0,0 +1,57 @@
...
+FROM openvino/model_server:2021.4.1-gpu
+USER root 
+
+RUN apt-get update && \
+    dpkg --get-selections | grep -v deinstall | awk '{print $1}' > base_packages.txt  && \
+    apt-get install -y --no-install-recommends ${DEPS} && \
+    rm -rf /var/lib/apt/lists/*
+    
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    build-essential \
```
mzegla, line 12: Since you removed `build-essential` from one Dockerfile, maybe you could also remove it here?

By the way, with removing `build-essential` from the image, wouldn't `third-party-programs.txt` with licenses get smaller?

DrewBearly: thanks, also removed here - we're keeping the third party mega-license in as is because I've been told we need to reflect packages in the base dockerfile - at least what our open source approval wanted us to do.

DrewBearly:
@ryanloney Just want to keep this alive - is there anything else preventing merging?

ryanloney (approved):

ryanloney:
@DrewBearly I have no further concerns except that the technical owners of the openvino_contrib repo approve this. 

@alvoron - can you let us know if you have any concerns? This is an integration into a Microsoft Azure product that uses OpenVINO and OpenVINO Model Server.

alvoron (approved):

PR #387, opened 2022-06-30
Do not trigger Azure CI on optimum changes

helena-intel:


helena-intel:
Merging to test functionality. Only affects Azure Linux CI.

PR #431, opened 2022-09-14
[Java] Add method release() for InferRequest for better resources management

dkurt:
I've faced a bug with ARM plugin when InferRequest produces correct output just once. Subsequent `infer()` calls return same output for different inputs. So I have recreate InferRequest for every new data. Method `release()` helps free memory before garbage collector.

### DIFF
--- a/modules/java_api/org/intel/openvino/InferRequest.java
+++ b/modules/java_api/org/intel/openvino/InferRequest.java
@@ -4,6 +4,8 @@
 package org.intel.openvino;
 
 public class InferRequest extends Wrapper {
+    private boolean isReleased = false;
+
     protected InferRequest(long addr) {
         super(addr);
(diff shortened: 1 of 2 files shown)


alvoron:
@dkurt thank you for you support of contrib modules! 

@likholat could you please review the changes?

modules/java_api/org/intel/openvino/InferRequest.java
```diff
@@ -76,6 +78,11 @@ public Tensor get_tensor(String tensorName) {
         return new Tensor(GetTensor(nativeObj, tensorName));
     }
 
+    public void release() {
```
likholat, line 81: Could you please add documentation for this method

likholat (approved):

PR #458, opened 2022-10-24
[IE TESTS] add definition set_device_name for cuda unit tests

iefode:


iefode:
@lohika-denis-kotov Please review & merge the changes.

FYI: We have not to link shared functional tests lib to unit tests binary. It's different testing level: unit is for class entity level (internal & etc), functional is for public API level testing. https://github.com/openvinotoolkit/openvino/blob/master/src/tests/README.md

lohika-denis-kotov:
> @lohika-denis-kotov Please review & merge the changes.
> 

@iefode Why do you add this function also for unit tests ?

> FYI: We have not to link shared functional tests lib to unit tests binary. It's different testing level: unit is for class entity level (internal & etc), functional is for public API level testing. https://github.com/openvinotoolkit/openvino/blob/master/src/tests/README.md

It probably was done by mistake, simple copy-paste ... Dependency should be tried to remove and see if it compiles

iefode:
> @iefode
> 
> > @lohika-denis-kotov Please review & merge the changes.
> 
> Why do you add this function also for unit tests ?
> 
> > FYI: We have not to link shared functional tests lib to unit tests binary. It's different testing level: unit is for class entity level (internal & etc), functional is for public API level testing. https://github.com/openvinotoolkit/openvino/blob/master/src/tests/README.md
> 
> It probably was done by mistake, simple copy-paste ... Dependency should be tried to remove and see if it compiles

It was my point. I don't understand why the shared lib was linked to the unit tests. But anyway I have a linking problem in the OV PR ( https://github.com/openvinotoolkit/openvino/pull/13340)

PR #462, opened 2022-11-03
Remove `Parser/pgen`

p-wysocki:
`Parser/pgen` is not included in Python 3.10 (it is removed from CPython). Having it fails building Python 3.10.


jiwaszki (approved):
🆗 👍

allnes:
@p-wysocki  hello!

This PR broke mac arm64 build in post-commit:

https://github.com/openvinotoolkit/openvino_contrib/runs/9284165171

p-wysocki:
@allnes Hi! 
The post-commit check is running on Python 3.11 - OpenVINO does not support it (we are currently enabling Python 3.10 support), and neither does `pybind` (the package where the errors are coming from). There's just an initial support for 3.11 (https://github.com/pybind/pybind11/releases/tag/v2.10.0).

PR #466, opened 2022-11-09
Update openvino-optimum requirements in setup.py

helena-intel:
Limit transformers version (to same upper version as .[nncf] install already has), add protobuf and sentencepiece to requirements to enable more models by default.


mzegla (approved):

helena-intel:
All optimum tests passed, merged without waiting for unrelated tests to pass.

PR #481, opened 2022-11-28
OpenVINO CPU extensions: custom operations support

likholat:
This module:
1. is a tutorial of how to support custom CPU operations, unsupported by OpenVINO (how to create OpenVINO CPU extensions)
2. includes some custom layers that are still required for open source projects that use OpenVINO (see the list in the README)

### DIFF
--- a/.ci/azure/linux.yml
+++ b/.ci/azure/linux.yml
@@ -49,6 +49,7 @@ jobs:
     BIN_DIR: $(OPENVINO_REPO_DIR)/bin/intel64/$(BUILD_TYPE)
     INSTALL_DIR: $(WORK_DIR)/install_pkg
     SETUPVARS: $(INSTALL_DIR)/setupvars.sh
+    CUSTOM_OP_LIB: $(BIN_DIR)/libuser_ov_extensions.so
     GRADLE_VER: 7.1.1
 
   steps:
@@ -93,7 +94,7 @@ jobs:
       set -e
       sudo -E $(OPENVINO_REPO_DIR)/install_build_dependencies.sh
       # Move jdk into contrib install_build_dependencies.sh
-      sudo apt --assume-yes install default-jdk
+      sudo apt --assume-yes install default-jdk libopencv-dev
       # Install gradle for Java
       sudo apt -y --no-install-recommends install unzip
       wget https://services.gradle.org/distributions/gradle-$(GRADLE_VER)-bin.zip
@@ -155,3 +156,20 @@ jobs:
       done
     workingDirectory: $(REPO_DIR)/modules/java_api
     displayName: 'Java tests'
+
+  - script: |
+      python3 -m pip install --user virtualenv
+      python3 -m virtualenv -p /usr/bin/python3.8 .env3
+      source .env3/bin/activate
+      python -m pip install --upgrade pip
+      python -m pip install -r $(REPO_DIR)/modules/custom_operations/tests/requirements.txt
+      cd ${OPENVINO_REPO_DIR}/tools && python -m pip install mo/
+    workingDirectory: $(WORK_DIR)
+    displayName: 'Create custom ops env'
+
+  - script: |
+      . $(SETUPVARS)
+      source  $(WORK_DIR)/.env3/bin/activate
+      python -m pytest tests/run_tests.py
+    workingDirectory: $(REPO_DIR)/modules/custom_operations
+    displayName: 'Custom ops tests'
--- a/README.md
+++ b/README.md
@@ -12,6 +12,7 @@ This list gives an overview of all modules available inside the contrib reposito
 * [**nvidia_plugin**](./modules/nvidia_plugin): NVIDIA GPU Plugin -- allows to perform deep neural networks inference on NVIDIA GPUs using CUDA, using OpenVINO API.
 * [**java_api**](./modules/java_api): Inference Engine Java API -- provides Java wrappers for Inference Engine public API.
 * [**Azure Video Analyzer**](./modules/ovms_ai_extension/): Azure Video Analyzer Extension -- enables exchange of video frames and inference results between [Azure Video Analyzer (AVA)](https://docs.microsoft.com/en-us/azure/azure-video-analyzer/video-analyzer-docs/overview) and OpenVINO™ Model Server.
+* [**custom_operations**](./modules/custom_operations/): Collection of Custom Operations -- implement Custom Operations with OpenVINO Extensibility Mechanism.
 
 ## How to build OpenVINO with extra modules
 You can build OpenVINO, so it will include the modules from this repository. Contrib modules are under constant development and it is recommended to use them alongside the master branch or latest releases of OpenVINO.
@@ -35,6 +36,7 @@ Additional build instructions are available for the following modules:
 * [**arm_plugin**](./modules/arm_plugin/README.md)
 * [**nvidia_plugin**](./modules/nvidia_plugin/README.md)
 * [**ovms_ai_extension**](./modules/ovms_ai_extension/README.md)
+* [**custom_operations**](./modules/custom_operations/README.md)
 
 ## Update the repository documentation
 In order to keep a clean overview containing all contributed modules, the following files need to be created/adapted:
--- a/modules/custom_operations/README.md
+++ b/modules/custom_operations/README.md
@@ -1,0 +1,66 @@
+# Collection of Custom Operations using OpenVINO Extensibility Mechanism
+
+This module provides a guide and the implementation of a few custom operations in the Intel OpenVINO runtime using its [Extensibility Mechanism](https://docs.openvino.ai/latest/openvino_docs_Extensibility_UG_Intro.html).
+
+There are some use cases when OpenVINO Custom Operations could be applicable:
+
+* There is an ONNX model which contains an operation not supported by OpenVINO.
+* You have a PyTorch model, which could be converted to ONNX, with an operation not supported by OpenVINO.
+* You want to replace a subgraph for ONNX model with one custom operation which would be supported by OpenVINO.
+
+More specifically, here we implement custom OpenVINO operations that add support for the following native PyTorch operation:
+
+* [torch.fft](examples/fft)
+
+And other custom operations introduced by third-party frameworks:
+
+* [calculate_grid](/examples/calculate_grid) and [sparse_conv](/examples/sparse_conv) from [Open3D](https://github.com/isl-org/Open3D)
+* [complex_mul](/examples/complex_mul) from [DIRECT](https://github.com/NKI-AI/direct)
+
+You can find more information about how to create and use OpenVINO Extensions to facilitate mapping of custom operations from framework model representation to OpenVINO representation [here](https://docs.openvino.ai/latest/openvino_docs_Extensibility_UG_Frontend_Extensions.html).
+
+
+## Build custom OpenVINO operation extension library
+
+The C++ code implementing the custom operation is in the `user_ie_extensions` directory. You'll have to build an "extension library" from this code so that it can be loaded at runtime. The steps below describe the build process:
+
+1. Install [OpenVINO Runtime for C++](https://docs.openvino.ai/latest/openvino_docs_install_guides_install_dev_tools.html#for-c-developers).
+
+2. Build the library:
+
+```bash
+cd user_ie_extensions
+mkdir build && cd build
+cmake .. -DCMAKE_BUILD_TYPE=Release && cmake --build . --parallel 4
+```
+
+If you need to build only some operations specify them with the `-DCUSTOM_OPERATIONS` option:
+```bash
+cmake .. -DCMAKE_BUILD_TYPE=Release -DCUSTOM_OPERATIONS=complex_mul\;fft
+```
+
+- Please note that [OpenCV](https://opencv.org/) installation is required to build an extension for the [fft](examples/fft) operation. Other extentions still can be built without OpenCV.
+
+You also could build the extension library [while building OpenVINO](../../README.md).
+
+## Load and use custom OpenVINO operation extension library
+
+You can use the custom OpenVINO operations implementation by loading it into the OpenVINO `Core` object at runtime. Then, load the model from the ONNX file with the `read_model()` API. Here's how to do that in Python:
+
+```python
+from openvino.runtime import Core
+
+# Create Core and register user extension
+core = Core()
+core.add_extension('/path/to/libuser_ov_extensions.so')
+
+# Load model from .onnx file directly
+model = core.read_model('model.onnx')
+compiled_model = core.compile_model(model, 'CPU')
+```
+
+You also can get OpenVINO IR model with Model Optimizer, just use extra `--extension` flag to specify a path to custom extensions:
+
+```bash
+mo --input_model model.onnx --extension /path/to/libuser_ov_extensions.so
+```
--- a/modules/custom_operations/examples/calculate_grid/export_model.py
+++ b/modules/custom_operations/examples/calculate_grid/export_model.py
@@ -1,0 +1,51 @@
+# Copyright (C) 2018-2022 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+import numpy as np
+import argparse
+import torch
+import torch.nn as nn
+from torch.autograd import Variable
+from .calculate_grid import CalculateGrid
+
+
+class MyModel(nn.Module):
+    def __init__(self):
+        super(MyModel, self).__init__()
+        self.calculate_grid = CalculateGrid()
+
+    def forward(self, x):
+        return self.calculate_grid.apply(x)
+
+
+def export(num_points, max_grid_extent):
+    # Generate a list of unique positions and add a mantissa
+    np.random.seed(32)
+    torch.manual_seed(11)
+
+    inp_pos = np.random.randint(0, max_grid_extent, [num_points, 3])
+    inp_pos = torch.tensor(inp_pos) + torch.rand(inp_pos.shape, dtype=torch.float32) # [0, 1)
+
+    model = MyModel()
+    with torch.no_grad():
+        torch.onnx.export(model, (inp_pos), 'model.onnx',
+                          input_names=['input'],
+                          output_names=['output'],
+                          operator_export_type=torch.onnx.OperatorExportTypes.ONNX_ATEN_FALLBACK)
+
+    ref = model(inp_pos).detach().numpy()
+
+    # Pad with sentinel values (-1, 0, 0) and zeros
+    ref = np.concatenate((ref, [[-1, 0, 0]]))
+    ref = np.pad(ref, ((0, inp_pos.shape[0] - ref.shape[0]), (0, 0)))
+
+    return [inp_pos.detach().numpy()], ref
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description='Generate ONNX model and test data')
+    parser.add_argument('--num_points', type=int, default=10)
+    parser.add_argument('--max_grid_extent', type=int, default=5)
+    args = parser.parse_args()
+
+    export(args.num_points, args.max_grid_extent)
--- a/modules/custom_operations/examples/fft/export_model.py
+++ b/modules/custom_operations/examples/fft/export_model.py
@@ -1,0 +1,49 @@
+# Copyright (C) 2018-2022 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+import numpy as np
+import argparse
+import torch
+import torch.nn as nn
+from torch.autograd import Variable
+from .fft import FFT
+
+
+class MyModel(nn.Module):
+    def __init__(self, inverse, centered, dims):
+        super(MyModel, self).__init__()
+        self.inverse = inverse
+        self.centered = centered
+        self.dims = dims
+        self.fft = FFT()
+
+    def forward(self, x):
+        return self.fft.apply(x, self.inverse, self.centered, self.dims)
+
+
+def export(shape, inverse, centered, dims):
+    np.random.seed(324)
+    torch.manual_seed(32)
+
+    model = MyModel(inverse, centered, dims)
+    inp = Variable(torch.randn(shape))
+    model.eval()
+
+    with torch.no_grad():
+        torch.onnx.export(model, inp, 'model.onnx',
+                          input_names=['input'],
+                          output_names=['output'],
+                          operator_export_type=torch.onnx.OperatorExportTypes.ONNX_FALLTHROUGH)
+
+    ref = model(inp)
+    return [inp.detach().numpy()], ref.detach().numpy()
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(description='Generate ONNX model and test data')
+    parser.add_argument('--shape', type=int, nargs='+', default=[5, 3, 6, 8, 2])
+    parser.add_argument('--inverse', action='store_true')
+    parser.add_argument('--centered', action='store_true')
+    parser.add_argument('--dims', type=int, nargs='+', default=[2, 3])
+    args = parser.parse_args()
+    export(args.shape, args.inverse, args.centered, args.dims)
--- a/modules/custom_operations/examples/sparse_conv/export_model.py
+++ b/modules/custom_operations/examples/sparse_conv/export_model.py
@@ -1,0 +1,72 @@
+# Copyright (C) 2018-2022 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+import numpy as np
+import argparse
+import torch
+import torch.nn as nn
+from torch.autograd import Variable
+from .sparse_conv import SparseConvONNX, SparseConvTransposeONNX
+
+
+def export(num_inp_points, num_out_points, max_grid_extent,
+           in_channels, filters, kernel_size, transpose):
+    np.random.seed(324)
+    torch.manual_seed(32)
+
+    if transpose:
+        sparse_conv = SparseConvTransposeONNX(in_channels=in_channels,
+                                              filters=filters,
+                                              kernel_size=kernel_size,
+                                              use_bias=False,
+                                              normalize=False)
+    else:
+        sparse_conv = SparseConvONNX(in_channels=in_channels,
+                                     filters=filters,
+                                     kernel_size=kernel_size,
+                                     use_bias=False,
+                                     normalize=False)
+
+    # Generate a list of unique positions and add a mantissa
+    def gen_pos(num_points):
+        inp_pos = np.random.randint(0, max_grid_extent, [num_points, 3])
+        inp_pos = np.unique(inp_pos, axis=0).astype(np.float32)
+        inp_pos = torch.tensor(inp_pos) + torch.rand(inp_pos.shape, dtype=torch.float32) # [0, 1)
+        return inp_pos
+
+    inp_pos = gen_pos(num_inp_points)
+    out_pos = gen_pos(num_out_points) if num_out_points else inp_pos
+
+    features = torch.randn([inp_pos.shape[0], in_channels])
+
+    voxel_size = torch.tensor(1.0)
+    sparse_conv.eval()
+
+    new_kernel = torch.randn(sparse_conv.state_dict()["kernel"].shape)
+    sparse_conv.load_state_dict({"kernel": new_kernel,
+                                 "offset": sparse_conv.state_dict()["offset"]})
+
+    with torch.no_grad():
+        torch.onnx.export(sparse_conv, (features, inp_pos, out_pos, voxel_size), 'model.onnx',
+                          input_names=['input', 'input1', 'input2', 'voxel_size'],
+                          output_names=['output'],
+                          operator_export_type=torch.onnx.OperatorExportTypes.ONNX_ATEN_FALLBACK)
+
+    ref = sparse_conv(features, inp_pos, out_pos, voxel_size)
+    return [features.detach().numpy(), inp_pos.detach().numpy(),
+            out_pos.detach().numpy()], ref.detach().numpy()
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description='Generate ONNX model and test data')
+    parser.add_argument('--num_inp_points', type=int)
+    parser.add_argument('--num_out_points', type=int)
+    parser.add_argument('--max_grid_extent', type=int)
+    parser.add_argument('--in_channels', type=int)
+    parser.add_argument('--filters', type=int)
+    parser.add_argument('--kernel_size', type=int, nargs='+')
+    parser.add_argument('--transpose', action='store_true')
+    args = parser.parse_args()
+
+    export(args.num_inp_points, args.num_out_points, args.max_grid_extent,
+           args.in_channels, args.filters, args.kernel_size, args.transpose)
--- a/modules/custom_operations/tests/requirements.txt
+++ b/modules/custom_operations/tests/requirements.txt
@@ -1,0 +1,4 @@
+torch==1.8.1
+open3d==0.14.1
+tensorboard
+pytest
--- a/modules/custom_operations/tests/run_tests.py
+++ b/modules/custom_operations/tests/run_tests.py
@@ -1,0 +1,94 @@
+# Copyright (C) 2018-2022 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+from openvino.runtime import Core
+from openvino.tools.mo import convert_model
+
+import subprocess
+import pytest
+from pathlib import Path
+import numpy as np
+import os
+
+
+def run_test(ref_inputs, ref_res, test_onnx=False, threshold=1e-5): 
+    inputs = {}
+    shapes = {}
+    for i in range(len(ref_inputs)):
+        suffix = '{}'.format(i if i > 0 else '')
+        inputs['input' + suffix] = ref_inputs[i]
+        shapes['input' + suffix] = ref_inputs[i].shape
+
+    ext_path = os.getenv('CUSTOM_OP_LIB')
+
+    core = Core()
+    core.add_extension(ext_path)
+
+    net = core.read_model('model.onnx') if test_onnx else convert_model('model.onnx', extensions=ext_path)
+
+    net.reshape(shapes)
+    exec_net = core.compile_model(net, 'CPU')
+
+    out = exec_net.infer_new_request(inputs)
+    out = next(iter(out.values()))
+
+    assert ref_res.shape == out.shape
+    diff = np.max(np.abs(ref_res - out))
+    assert diff <= threshold
+
+
+@pytest.mark.parametrize("shape", [[5, 120, 2], [4, 240, 320, 2], [3, 16, 240, 320, 2], [4, 5, 16, 31, 2]])
+@pytest.mark.parametrize("inverse", [False, True])
+@pytest.mark.parametrize("centered", [False, True])
+@pytest.mark.parametrize("test_onnx", [False, True])
+@pytest.mark.parametrize("dims", [[1], [1, 2], [2, 3]])
+def test_fft(shape, inverse, centered, test_onnx, dims):
+    from examples.fft.export_model import export
+
+    if len(shape) == 3 and dims != [1] or \
+       len(shape) == 4 and dims == [2, 3] or \
+       len(shape) == 5 and dims == [1] or \
+       centered and len(dims) != 2:
+        pytest.skip("unsupported configuration")
+
+    inp, ref = export(shape, inverse, centered, dims)
+    run_test(inp, ref, test_onnx=test_onnx)
+
+
+@pytest.mark.parametrize("shape", [[3, 2, 4, 8, 2], [3, 1, 4, 8, 2]])
+@pytest.mark.parametrize("test_onnx", [False, True])
+def test_complex_mul(shape, test_onnx):
+    from examples.complex_mul.export_model import export
+
+    inp, ref = export(other_shape=shape)
+    run_test(inp, ref, test_onnx=test_onnx)
+
+
+@pytest.mark.parametrize("in_channels", [1, 3])
+@pytest.mark.parametrize("filters", [1, 4])
+@pytest.mark.parametrize("kernel_size", [[3, 3, 3], [5, 5, 5], [2, 2, 2]])
+@pytest.mark.parametrize("out_pos", [None, 16])
+def test_sparse_conv(in_channels, filters, kernel_size, out_pos):
+    from examples.sparse_conv.export_model import export
+
+    inp, ref = export(num_inp_points=1000, num_out_points=out_pos, max_grid_extent=4, in_channels=in_channels,
+                    filters=filters, kernel_size=kernel_size, transpose=False)
+    run_test(inp, ref, test_onnx=True, threshold=1e-4)
+
+
+@pytest.mark.parametrize("in_channels", [1, 3])
+@pytest.mark.parametrize("filters", [1, 4])
+@pytest.mark.parametrize("kernel_size", [[3, 3, 3], [5, 5, 5]])
+@pytest.mark.parametrize("out_pos", [None, 16])
+def test_sparse_conv_transpose(in_channels, filters, kernel_size, out_pos):
+    from examples.sparse_conv.export_model import export
+
+    inp, ref = export(num_inp_points=1000, num_out_points=out_pos, max_grid_extent=4, in_channels=in_channels,
+                    filters=filters, kernel_size=kernel_size, transpose=True)
+    run_test(inp, ref, test_onnx=True, threshold=1e-4)
+
+
+def test_calculate_grid():
+    from examples.calculate_grid.export_model import export
+    inp, ref = export(num_points=10, max_grid_extent=5)
+    run_test(inp, ref, test_onnx=True)
--- a/modules/custom_operations/user_ie_extensions/CMakeLists.txt
+++ b/modules/custom_operations/user_ie_extensions/CMakeLists.txt
@@ -1,0 +1,64 @@
+# Copyright (C) 2022 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+#
+
+set(CMAKE_CXX_STANDARD 11)
+
+set(TARGET_NAME "user_ov_extensions")
+
+find_package(OpenVINO REQUIRED)
+find_package(TBB COMPONENTS tbb tbbmalloc)
+find_package(OpenCV COMPONENTS core)
+
+set(OP_REQ_TBB "complex_mul" "fft")
+
+if(CUSTOM_OPERATIONS)
+  file(GLOB SRC ov_extension.cpp)
+
+  foreach(op IN LISTS CUSTOM_OPERATIONS)
+    list(APPEND SRC ${CMAKE_CURRENT_SOURCE_DIR}/${op}.cpp)
+  endforeach()
+else()
+  file(GLOB SRC *.cpp)
+
+  foreach(op IN LISTS SRC)
+    get_filename_component(op_name "${op}" NAME_WLE)
+    list(APPEND CUSTOM_OPERATIONS ${op_name})
+  endforeach()
+
+  list(REMOVE_ITEM CUSTOM_OPERATIONS ov_extension)
+endif()
+
+if (NOT OpenCV_FOUND)
+  list(REMOVE_ITEM SRC ${CMAKE_CURRENT_SOURCE_DIR}/fft.cpp)
+  list(REMOVE_ITEM CUSTOM_OPERATIONS fft)
+endif()
+
+if (NOT TBB_FOUND)
+  foreach(op IN LISTS OP_REQ_TBB)
+    list(REMOVE_ITEM SRC ${CMAKE_CURRENT_SOURCE_DIR}/${op}.cpp)
+    list(REMOVE_ITEM CUSTOM_OPERATIONS ${op})
+  endforeach()
+endif()
+
+add_library(${TARGET_NAME} SHARED ${SRC})
+
+if (OpenCV_FOUND AND "fft" IN_LIST CUSTOM_OPERATIONS)
+  target_link_libraries(${TARGET_NAME} PRIVATE ${OpenCV_LIBRARIES})
+endif()
+
+if (TBB_FOUND)
+  foreach(op IN LISTS OP_REQ_TBB)
+    if(op IN_LIST CUSTOM_OPERATIONS)
+      set(TBB_REQUIRED TRUE)
+    endif()
+  endforeach()
+
+  if(TBB_REQUIRED)
+    target_link_libraries(${TARGET_NAME} PRIVATE TBB::tbb TBB::tbbmalloc)
+  endif()
+endif()
+
+target_link_libraries(${TARGET_NAME} PRIVATE openvino::runtime)
+
+target_compile_definitions(${TARGET_NAME} PRIVATE ${CUSTOM_OPERATIONS})
--- a/modules/custom_operations/user_ie_extensions/calculate_grid.cpp
+++ b/modules/custom_operations/user_ie_extensions/calculate_grid.cpp
@@ -1,0 +1,65 @@
+// Copyright (C) 2018-2022 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#include "calculate_grid.hpp"
+
+using namespace TemplateExtension;
+
+CalculateGrid::CalculateGrid(const ov::Output<ov::Node>& inp_pos) : Op({inp_pos}) {
+    constructor_validate_and_infer_types();
+}
+
+void CalculateGrid::validate_and_infer_types() {
+    auto outShape = get_input_partial_shape(0);
+    set_output_type(0, get_input_element_type(0), outShape);
+}
+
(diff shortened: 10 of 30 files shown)


likholat:
@dkurt, please, take a look

modules/openvino_pytorch_layers/README.md
```diff
@@ -0,0 +1,46 @@
+Module with guides how to enable in Intel OpenVINO some layers from PyTorch:
```
dkurt, line 1: Maybe name module just by `custom_layers`?

modules/openvino_pytorch_layers/README.md
```diff
@@ -0,0 +1,46 @@
...
+
```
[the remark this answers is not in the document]
likholat: @ngaloppo and I had a conversation and we decided not to publish PyPI package

modules/openvino_pytorch_layers/examples/calculate_grid/export_model.py
```diff
@@ -0,0 +1,40 @@
...
+                          operator_export_type=torch.onnx.OperatorExportTypes.ONNX_ATEN_FALLBACK)
+
+    ref = model(inp_pos).detach().numpy()
+
+    # Pad values with espetial end line (-1, 0, 0) and zeros
+    ref = np.concatenate((ref, [[-1, 0, 0]]))
+    ref = np.pad(ref, ((0, inp_pos.shape[0] - ref.shape[0]), (0, 0)))
+
+    np.save('inp', inp_pos.detach().numpy())
+    np.save('ref', ref)
```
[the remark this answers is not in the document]
likholat: Thanks, done!

modules/openvino_pytorch_layers/README.md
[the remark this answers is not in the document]
ngaloppo: For now :) My concern is mostly about where to publish, and license requirements.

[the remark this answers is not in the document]
likholat: @ngaloppo @dkurt, now we need only one `.so` file to use custom layers for MO or IE
Does it make sense to create package with one `.so` file?

modules/custom_layers/README.md
```diff
@@ -0,0 +1,46 @@
+Module with guides how to enable in Intel OpenVINO some layers from PyTorch:
```
ngaloppo, line 1:
```suggestion
# Collection of Custom Operations using OpenVINO Extensibility Mechanism

This module provides a guide and the implementation a few custom operations in the Intel OpenVINO runtime using its [Extensibility Mechanism](https://docs.openvino.ai/latest/openvino_docs_Extensibility_UG_Intro.html). 

More specifically, here we implement custom OpenVINO operations that add support for the following native PyTorch operations:
```

modules/custom_layers/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+* [torch.fft](examples/fft)
+* [nn.functional.grid_sample](/examples/grid_sample)
+
+And other custom operations:
```
ngaloppo, line 6:
```suggestion
And other custom operations implemented by third-party frameworks:
```

modules/custom_layers/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+* [calculate_grid](/examples/calculate_grid) and [sparse_conv](/examples/sparse_conv) from [Open3D](https://github.com/isl-org/Open3D)
+* [complex_mul](/examples/complex_mul) from [DIRECT](https://github.com/NKI-AI/direct)
+
+
+## OpenVINO custom CPU extensions
+
+You  need to build CPU extensions library which actually has C++ layers implementations:
```
ngaloppo, line 14:
```suggestion
## Build custom OpenVINO operation extension library

The C++ code implementing the custom operation is in the `user_ie_extensions` directory. You'll have to build an "extension library" from this code so that it can be loaded at runtime. The steps below describe the build process:
```

modules/custom_layers/README.md
```diff
@@ -0,0 +1,46 @@
...
+```python
+from openvino.runtime import Core
+
+# Create Core and register user extension
+core = Core()
+core.add_extension('/path/to/user_ie_extensions/build/libuser_cpu_extension.so')
+
+# Load model from .onnx file directly
+model = ie.read_model('model.onnx')
+compiled_model = ie.compile_model(model, 'CPU')
```
ngaloppo, line 34:
```suggestion
model = core.read_model('model.onnx')
compiled_model = core.compile_model(model, 'CPU')
```

modules/custom_layers/README.md
```diff
@@ -0,0 +1,46 @@
...
+You  need to build CPU extensions library which actually has C++ layers implementations:
+```bash
+source /opt/intel/openvino_2022/setupvars.sh
+
+cd user_ie_extensions
+mkdir build && cd build
+cmake .. -DCMAKE_BUILD_TYPE=Release && make -j$(nproc --all)
+```
+
+And add compiled extensions library to your project:
```
ngaloppo, line 23:
```suggestion
## Load and use custom OpenVINO operation extension library

You can use the custom OpenVINO operations implementation by loading it into the OpenVINO `Core` object at runtime. Then, load the model from the ONNX file with the `read_model()` API. Here's how to do that in Python:
```

modules/custom_layers/README.md
```diff
@@ -0,0 +1,46 @@
...
+# Create Core and register user extension
+core = Core()
+core.add_extension('/path/to/user_ie_extensions/build/libuser_cpu_extension.so')
+
+# Load model from .onnx file directly
+model = ie.read_model('model.onnx')
+compiled_model = ie.compile_model(model, 'CPU')
+```
+
+- You also can get OpenVINO IR model with Model Optimizer, just use extra `--extension` flag to specify a path to custom extensions:
```
ngaloppo, line 37:
```suggestion
You also can get OpenVINO IR model with Model Optimizer, just use extra `--extension` flag to specify a path to custom extensions:
```

modules/custom_layers/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+```bash
+mo --input_model model.onnx --extension /path/to/user_ie_extensions/build/libuser_cpu_extension.so
+```
+
+## These layers were supported for:
+* [/examples/complex_mul](/examples/complex_mul) and [examples/fft](examples/fft) for [DIRECT](https://github.com/NKI-AI/direct)
+* [nn.functional.grid_sample](/examples/grid_sample) for [RAFT](https://github.com/princeton-vl/RAFT)
+* [calculate_grid](/examples/calculate_grid) and [sparse_conv](/examples/sparse_conv) for [Open3D](https://github.com/isl-org/Open3D)
```
ngaloppo, line 46: Is this a repeat of the information at the start of the README?

ngaloppo (changes requested):

modules/openvino_pytorch_layers/README.md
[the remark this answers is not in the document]
dkurt: @likholat, I agree, due now a single .so is used as extension, a wheel can be used just for publishing.

modules/custom_layers/README.md
dkurt: Maybe introduced by third-party frameworks? Because they were re-implemented from scratch.

modules/custom_layers/README.md
likholat: There are more projects mentioned here than at the beginning, because first I briefly describe each layer, and here I write what projects they were created for. But yes, the information is repeated.

modules/custom_layers/README.md
ngaloppo: Good catch!

modules/custom_layers/README.md
ngaloppo: I guess we can just remove this paragraph then...

ngaloppo (changes requested):

ngaloppo:
Should we call the module `custom_ops` or `custom_operations` instead of `custom_layers`? We've been removing the term "layers" from the OpenVINO nomenclature.

.ci/azure/linux.yml
```diff
@@ -159,6 +162,20 @@ jobs:
...
+  - script: |
+      . $(SETUPVARS)
+      cmake $(REPO_DIR)/modules/custom_operations/user_ie_extensions && make
+    workingDirectory: $(BUILD_DIR_CUSTOM_OP)
+    displayName: 'CMake Custom Layers'
+
+  - script: |
+      . $(SETUPVARS)
+      python3 -m pip install --upgrade pip
+      python3 -m pip install openvino-dev
```
dkurt, line 174: OpenVINO from the install folder should be used

.ci/azure/linux.yml
```diff
@@ -159,6 +162,20 @@ jobs:
     workingDirectory: $(REPO_DIR)/modules/java_api
     displayName: 'Java tests'
 
+  - script: |
+      . $(SETUPVARS)
+      cmake $(REPO_DIR)/modules/custom_operations/user_ie_extensions && make
```
dkurt, line 167:
```
cmake -DCMAKE_BUILD_TYPE=Release $(REPO_DIR)/modules/custom_operations/user_ie_extensions && make -j$(nproc --all)
```

likholat:
@dkurt , @ngaloppo , thanks for the review!:blush: I've fixed all the comments and I think, now it's ready to merge

ngaloppo:
LGTM. @mbencer could you review with respect to the new extensibility API?

ngaloppo (approved):

modules/custom_operations/README.md
```diff
@@ -0,0 +1,48 @@
...
+## Load and use custom OpenVINO operation extension library
+
+You can use the custom OpenVINO operations implementation by loading it into the OpenVINO `Core` object at runtime. Then, load the model from the ONNX file with the `read_model()` API. Here's how to do that in Python:
+
+```python
+from openvino.runtime import Core
+
+# Create Core and register user extension
+core = Core()
+core.add_extension('/path/to/user_ie_extensions/build/libuser_cpu_extension.so')
```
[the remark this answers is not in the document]
likholat: Thanks, done!

modules/custom_operations/README.md
```diff
@@ -0,0 +1,50 @@
...
+## Build custom OpenVINO operation extension library
+
+The C++ code implementing the custom operation is in the `user_ie_extensions` directory. You'll have to build an "extension library" from this code so that it can be loaded at runtime. The steps below describe the build process:
+
+```bash
+source /opt/intel/openvino_2022/setupvars.sh
+
+cd user_ie_extensions
+mkdir build && cd build
+cmake .. -DCMAKE_BUILD_TYPE=Release && make -j$(nproc --all)
```
ilya-lavrenov, line 27: 1. nproc does not work for macOS AFAIK
2. make is not available out of box on Windows. 

Can we just use `cmake --build .. --parallel 4` ?

modules/custom_operations/user_ie_extensions/fft.cpp
```diff
@@ -0,0 +1,369 @@
...
+    if (!loaded) {
+        loaded = true;
+        try {
+#ifdef _WIN32
+            so.reset(new InferenceEngine::details::SharedObjectLoader("opencv_core.dll"));
+#elif defined(__APPLE__)
+            so.reset(new InferenceEngine::details::SharedObjectLoader("libopencv_core.dylib"));
+#else
+            so.reset(new InferenceEngine::details::SharedObjectLoader("libopencv_core.so"));
+#endif
```
ilya-lavrenov, line 39: @ilyachur do you have suggestions on this? 
We will have to rework this place once we remove API 1.0.. Moreover, details namespace is not allowed for public users.

modules/custom_operations/user_ie_extensions/fft.cpp
```diff
@@ -0,0 +1,369 @@
...
+
+            cvReleaseMat(&inp);
+            cvReleaseMat(&out);
+        });
+    } else if (dims.size() == 4 && numSignalDims == 1 && signalDimsData[0] == 1) {
+        int rows = dims[1];
+        int cols = dims[2];
+
+        const int planeSize = rows;
+        InferenceEngine::parallel_for(batch * cols, [&](size_t d) {
```
ilya-lavrenov, line 323: CC @ilyachur 
Looks like we need to port this to API 2.0 to allow creation of optimized CPU kernels..

modules/custom_operations/README.md
ngaloppo: I agree, if `cmake --build` makes this easier across multiple OS platforms, let's use that. Just FYI `nproc` is available on macOS.

ilya-lavrenov: > Just FYI nproc is available on macOS.

It's not available out of box, only one you've installed `coreutils` from brew. Native way is to use `sysctl -n hw.ncpu`

ngaloppo: Ah! I did not know that, good to know! Thanks!

modules/custom_operations/user_ie_extensions/fft.cpp
likholat: @ilya-lavrenov could you please clarify what you mean

ilya-lavrenov: I mean currently you have to use API 1.0 because we don't have such helpers in 2.0
we (OV team) need to think whether we need to move such helpers to 2.0 folder.

modules/custom_operations/tests/run_tests.py
```diff
@@ -0,0 +1,110 @@
...
+import pytest
+from pathlib import Path
+import numpy as np
+import os
+
+def convert_model():
+    subprocess.run(['mo',
+                    '--input_model=model.onnx',
+                    '--extension', os.path.join(os.getenv('BUILD_DIR_CUSTOM_OP'), 'libuser_cpu_extension.so')],
+                    check=True)
```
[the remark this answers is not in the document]
likholat: Thanks, done!

modules/openvino_pytorch_layers/README.md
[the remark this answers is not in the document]
ilya-lavrenov: if we think about publishing, we can consider github actions for this

modules/custom_operations/README.md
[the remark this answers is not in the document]
ilya-lavrenov: once this is merged, suggest to link to this module from OV docs with extensions.

modules/custom_operations/user_ie_extensions/fft.cpp
ilyachur: We discussed this topic multiple times and at the current moment we don't have a plan to move this functionality to new extensions.
In order to support custom op user can add evaluate method.
For performance execution he can contribute to CPU plugin

modules/custom_operations/examples/calculate_grid/export_model.py
```diff
@@ -0,0 +1,42 @@
...
+    model = MyModel()
+    with torch.no_grad():
+        torch.onnx.export(model, (inp_pos), 'model.onnx',
+                          input_names=['input'],
+                          output_names=['output'],
+                          operator_export_type=torch.onnx.OperatorExportTypes.ONNX_ATEN_FALLBACK)
+
+    ref = model(inp_pos).detach().numpy()
+
+    # Pad values with espetial end line (-1, 0, 0) and zeros
```
slyalin, line 38:
```suggestion
    # Pad values with espeсial end line (-1, 0, 0) and zeros
```

modules/custom_operations/user_ie_extensions/ov_extension.cpp
```diff
@@ -0,0 +1,37 @@
...
+        std::make_shared<ov::OpExtension<TemplateExtension::GridSample>>(),
+        std::make_shared<ov::OpExtension<TemplateExtension::FFT>>(),
+
+        // Register operaton mapping, required when converted from framework model format
+        std::make_shared<ov::frontend::OpExtension<TemplateExtension::SparseConv>>(),
+        std::make_shared<ov::frontend::OpExtension<TemplateExtension::SparseConvTranspose>>(),
+        std::make_shared<ov::frontend::OpExtension<TemplateExtension::ComplexMultiplication>>(),
+        std::make_shared<ov::frontend::OpExtension<TemplateExtension::CalculateGrid>>(),
+        std::make_shared<ov::frontend::OpExtension<TemplateExtension::GridSample>>(),
+        std::make_shared<ov::frontend::OpExtension<TemplateExtension::FFT>>(),
```
slyalin, line 34: @itikhono, this is what we want to replace by `OPENVINO_FAMEWORK_MAP` macro in the operation definition. Examples in this PR are all eligible for it. Could you refer to the code where usage examples can be found for the macro, and add the description of this macro to the documentation?

@likholat, do you by chance know about this macro? Have you tried to use it?

modules/custom_operations/user_ie_extensions/sparse_conv_transpose.cpp
```diff
@@ -0,0 +1,109 @@
...
+using namespace TemplateExtension;
+
+//! [op:ctor]
+SparseConvTranspose::SparseConvTranspose(const ov::Output<ov::Node>& features,
+                                         const ov::Output<ov::Node>& inp_pos,
+                                         const ov::Output<ov::Node>& out_pos,
+                                         const ov::Output<ov::Node>& kernel,
+                                         const ov::Output<ov::Node>& offset) : Op({features, inp_pos, out_pos, kernel, offset}) {
+    constructor_validate_and_infer_types();
+}
```
slyalin, line 16: For operations that we are not using manually in the code (with `std::make_shared<OpName>(params)`, we can save the number of code lines by defining ctor with vector of inputs and then just forward it to Op ctor. Then it will simplify the code in `clone_with_new_inputs`. Creating ctors with multiple arguments handy in case if you are manually building the graph of a model. Here we are not doing it.

modules/custom_operations/user_ie_extensions/sparse_conv_transpose.cpp
```diff
@@ -0,0 +1,109 @@
...
+    OPENVINO_ASSERT(new_args.size() == 5, "Incorrect number of new arguments");
+    return std::make_shared<SparseConvTranspose>(new_args.at(0), new_args.at(1), new_args.at(2), new_args.at(3), new_args.at(4));
+}
+//! [op:copy]
+
+//! [op:visit_attributes]
+bool SparseConvTranspose::visit_attributes(ov::AttributeVisitor& visitor) {
+    return true;
+}
+//! [op:visit_attributes]
```
slyalin, line 39: @ilyachur, we shouldn't require to implement this method if there are no attributes in an op. It makes more boilerplate code without any value. You can see, all ops in this PR don't have any attributes -- this is most likely true for all PT custom ops.

modules/custom_operations/tests/requirements.txt
```diff
@@ -0,0 +1,5 @@
+torch==1.8.1
+torchvision==0.9.1
+open3d==0.14.1
```
yury-gorbachev, line 3: Why is this needed?
Also, not sure about tensorboard

ilya-lavrenov:
can we also add links to main readme? and maybe Wiki

modules/custom_operations/examples/fft/export_model.py
```diff
@@ -0,0 +1,47 @@
...
+    ref = model(inp)
+    return [inp.detach().numpy()], ref.detach().numpy()
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(description='Generate ONNX model and test data')
+    parser.add_argument('--shape', type=int, nargs='+', default=[5, 3, 6, 8, 2])
+    args = parser.parse_args()
+
+    export(args.shape)
```
slyalin, line 47: Is it expected to work? `export` expects 4 arguments, provided one only. Doesn't work for me locally.

modules/custom_operations/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+The C++ code implementing the custom operation is in the `user_ie_extensions` directory. You'll have to build an "extension library" from this code so that it can be loaded at runtime. The steps below describe the build process:
+
+```bash
+source /opt/intel/openvino_2022/setupvars.sh
+
+cd user_ie_extensions
+mkdir build && cd build
+cmake .. -DCMAKE_BUILD_TYPE=Release && cmake --build . --parallel 4
+```
```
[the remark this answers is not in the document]
ilya-lavrenov: @likholat do you think we we explicitly list all source files here https://github.com/likholat/openvino_contrib/blob/custom_pytorch_layers/modules/custom_operations/user_ie_extensions/CMakeLists.txt#L17 and write in docs that users can keep only required source files to make a selective buid?

ilya-lavrenov:
After offline discussion agreed:
- Reflect new module in global Contrib's readme and Wiki
- Make OpenCV optional in cmake with warning on cmake configure level and reflect in README.md that OpenCV is required for particular operations.
- `loadOpenCV` seems to be unchecked in the FFT's ctor
- let's add root CMakeLists.txt file to enable automatic search via `-DOPENVINO_EXTRA_MODULES=<contrib>/modules`. Likely after this we don't need changes in `.ci` like `CMake Custom Layers`

modules/custom_operations/user_ie_extensions/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
...
+
+add_library(${TARGET_NAME} SHARED ${SRC})
+
+target_compile_definitions(${TARGET_NAME} PRIVATE IMPLEMENT_OPENVINO_EXTENSION_API)
+
+if (OpenCV_FOUND)
+  target_include_directories(${TARGET_NAME} PRIVATE ${OpenCV_INCLUDE_DIRS})
+endif()
+
+target_link_libraries(${TARGET_NAME} PRIVATE openvino::runtime ${TBB_IMPORTED_TARGETS})
```
ilya-lavrenov, line 27: latest TBB versions do not provide `TBB_IMPORTED_TARGETS` variable. Let's link explicitly with targets

modules/custom_operations/examples/calculate_grid/export_model.py
ngaloppo: @slyalin "special"? or `Pad with sentinel values (-1,0,0) and zeros`

modules/custom_operations/examples/fft/export_model.py
```diff
@@ -0,0 +1,47 @@
...
+class MyModel(nn.Module):
+    def __init__(self, inverse, centred, dims):
+        super(MyModel, self).__init__()
+        self.inverse = inverse
+        self.centred = centred
+        self.dims = dims
+        self.fft = FFT()
+
+    def forward(self, x):
+        return self.fft.apply(x, self.inverse, self.centred, self.dims)
```
ngaloppo, line 21:
```suggestion
    def __init__(self, inverse, centered, dims):
        super(MyModel, self).__init__()
        self.inverse = inverse
        self.centered = centered
        self.dims = dims
        self.fft = FFT()

    def forward(self, x):
        return self.fft.apply(x, self.inverse, self.centered, self.dims)
```

modules/custom_operations/tests/run_tests.py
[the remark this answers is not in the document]
ngaloppo: Love this!

modules/custom_operations/examples/fft/export_model.py
likholat: Thanks for noticing this! I think this is left from the previous implementation of the operation. Fixed.

modules/custom_operations/tests/requirements.txt
likholat: both `open3d` and `tensorboard` are needed to test (and show how to support) `SparseConv` operation
https://github.com/likholat/openvino_contrib/blob/197109ccff1c2dc9770e09ce85cc2a25754e6faf/modules/custom_operations/examples/sparse_conv/export_model.py#L9

modules/custom_operations/examples/sparse_conv/export_model.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description='Generate ONNX model and test data')
+    parser.add_argument('--num_inp_points', type=int)
+    parser.add_argument('--num_out_points', type=int)
+    parser.add_argument('--max_grid_extent', type=int)
+    parser.add_argument('--in_channels', type=int)
+    parser.add_argument('--filters', type=int)
+    parser.add_argument('--kernel_size', type=int)
```
slyalin, line 67: I cannot pass this value via command line -- it is `int` here, but later it is used as a list. In your tests this is a list as well. So this script cannot be used in command line.

modules/custom_operations/examples/calculate_grid/export_model.py
slyalin: Your variant is better, agree -- I just fixed grammar in the original word :)

modules/custom_operations/README.md
```diff
@@ -0,0 +1,50 @@
+# Collection of Custom Operations using OpenVINO Extensibility Mechanism
+
+This module provides a guide and the implementation a few custom operations in the Intel OpenVINO runtime using its [Extensibility Mechanism](https://docs.openvino.ai/latest/openvino_docs_Extensibility_UG_Intro.html).
+
```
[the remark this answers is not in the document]
slyalin: We haven't covered the extensibility story for Pytorch in the documentation. We have this document https://docs.openvino.ai/latest/openvino_docs_MO_DG_prepare_model_convert_model_Convert_Model_From_PyTorch.html that is basically a forward to `torch.onnx.export` tutorial in Pytorch documentation, and a couple of specific examples that are not illustrating custom ops.

As we haven't yet provided direct Pytorch importing then for the user this story breaks up into two steps: express it in ONNX first, and then import ONNX to the runtime. For the first part, we are not providing much help except mentioned `torch.onnx.export` link which is covering custom ops expression in ONNX.

With `openvino.tools.mo.convert_model` which hides `torch.onnx.export` underneath the story should be a bit more elegant, because we will try to export all ops in ONNX with `ONNX_FALLTHROUGH` regardless of support in ONNX default opset, and will be able to intercept custom ops in our ONNX FE as usual for other ONNX custom ops via extension. Currently, we are not doing that, but I manually tried to modify _a single line of code_ in the `convert_model` implementation and it worked for the `sparse_conv` exported model from this PR. So it is doable without explaining `torch.onnx.export` parameters and better to focus on `openvino.tools.mo.convert_model` polishing to avoid such details in our documentation.

modules/custom_operations/examples/calculate_grid/export_model.py
```diff
@@ -0,0 +1,42 @@
...
+    # Generate a list of unique positions and add a mantissa
+    np.random.seed(32)
+    torch.manual_seed(11)
+
+    inp_pos = np.random.randint(0, max_grid_extent, [num_points, 3])
+    inp_pos = torch.tensor(inp_pos) + torch.rand(inp_pos.shape, dtype=torch.float32) # [0, 1)
+
+    model = MyModel()
+    with torch.no_grad():
+        torch.onnx.export(model, (inp_pos), 'model.onnx',
```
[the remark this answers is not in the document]
slyalin: It should be as simple as `ov_model = convert_model(sparse_conv, example_input=(features, inp_pos, out_pos, voxel_size), extensions='libuser_cpu_extension.so')`. With the current state of OV master it doesn't work, because we are not passing `ONNX_FALLTHROUGH` by default inside `convert_model`, and for some reason I need to pass the full path to extension lib (even if it is placed in the current directory -- @ilyachur, @popovaan, is it expected?). Anyway with the already mentioned in another comment single line modification in `convert_model` this piece of code works with the current master and provides OpenVINO model with all extensions applied.

Note, that with `conver_model` we are not assigning names for inputs/outputs providing the capability to access them as positional arguments/results. Names of inputs/outputs look to be pure ONNX-export artifact here.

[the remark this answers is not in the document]
slyalin: Still it is not heavily tested, that's why we are not providing this extensibility story for users.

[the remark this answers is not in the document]
ilyachur: In general we don't provide a limitation for extensions loading, and it should be possible to load extension with using relative path.

modules/custom_operations/user_ie_extensions/calculate_grid.cpp
```diff
@@ -0,0 +1,79 @@
+// Copyright (C) 2018-2022 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#include "calculate_grid.hpp"
+
+using namespace TemplateExtension;
+
+//! [op:ctor]
```
ilyachur, line 9: I think you can remove these comments

modules/custom_operations/user_ie_extensions/calculate_grid.cpp
```diff
@@ -0,0 +1,79 @@
...
+    OPENVINO_ASSERT(new_args.size() == 1, "Incorrect number of new arguments");
+    return std::make_shared<CalculateGrid>(new_args.at(0));
+}
+//! [op:copy]
+
+//! [op:visit_attributes]
+bool CalculateGrid::visit_attributes(ov::AttributeVisitor& visitor) {
+    return true;
+}
+//! [op:visit_attributes]
```
ilyachur, line 33: This behaviour was changed for 2023.0 release you can remove it.
```suggestion
```

modules/custom_operations/user_ie_extensions/calculate_grid.cpp
```diff
@@ -0,0 +1,79 @@
...
+        out[i * 3 + 2] = 0.5f + std::get<2>(it);
+        i += 1;
+    }
+    memset(out + i * 3, 0, sizeof(float) * 3 * (numPoints - i));
+    out[i * 3] = -1.0f;
+    return true;
+}
+
+bool CalculateGrid::has_evaluate() const {
+    return true;
```
ilyachur, line 77: I think here you need to return `true` only for `f32` data type, because as I see other data types are not supported inside the evaluate method

modules/custom_operations/user_ie_extensions/calculate_grid.hpp
```diff
@@ -0,0 +1,29 @@
...
+
+class CalculateGrid : public ov::op::Op {
+public:
+    OPENVINO_OP("CalculateGrid");
+
+    CalculateGrid() = default;
+    CalculateGrid(const ov::Output<ov::Node>& inp_pos);
+    void validate_and_infer_types() override;
+    std::shared_ptr<ov::Node> clone_with_new_inputs(const ov::OutputVector& new_args) const override;
+    bool visit_attributes(ov::AttributeVisitor& visitor) override;
```
ilyachur, line 22:
```suggestion
```

modules/custom_operations/user_ie_extensions/complex_mul.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+    size_t channels0 = inputs[0].get_shape()[1];
+    size_t channels1 = inputs[1].get_shape()[1];
+    size_t batch = inputs[0].get_shape()[0];
+    size_t spatialSize = inputs[0].get_shape()[2] * inputs[0].get_shape()[3];
+
+    // x1 = x_r * y_r - x_i * y_i
+    // x2 = x_r * y_i + x_i * y_r
+    if (channels0 == channels1)
+        InferenceEngine::parallel_for(channels0 * batch, [&](size_t ch) {
```
ilyachur, line 51: After the merge of PR https://github.com/openvinotoolkit/openvino/pull/14485 you will be able to use:
```suggestion
        ov::parallel_for(channels0 * batch, [&](size_t ch) {
```

modules/custom_operations/user_ie_extensions/complex_mul.cpp
```diff
@@ -0,0 +1,85 @@
...
+            }
+        });
+    else
+        IE_THROW() << "Wrong number of channels for second input!";
+
+    return true;
+}
+
+bool ComplexMultiplication::has_evaluate() const {
+    return true;
```
ilyachur, line 83: Looks like only `f32` data type is supported

modules/custom_operations/user_ie_extensions/complex_mul.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+//! [op:copy]
+std::shared_ptr<ov::Node> ComplexMultiplication::clone_with_new_inputs(const ov::OutputVector& new_args) const {
+    OPENVINO_ASSERT(new_args.size() == 2, "Incorrect number of new arguments");
+    return std::make_shared<ComplexMultiplication>(new_args);
+}
+//! [op:copy]
+
+//! [op:visit_attributes]
+bool ComplexMultiplication::visit_attributes(ov::AttributeVisitor& visitor) {
```
ilyachur, line 32: this visit attributes also can be removed

modules/custom_operations/user_ie_extensions/fft.cpp
ilyachur: I am not sure that we need to provide an API to load some custom runtime libraries, it is not a responsibility of OpenVINO toolkit.

modules/custom_operations/user_ie_extensions/fft.cpp
```diff
@@ -0,0 +1,366 @@
...
+    OPENVINO_ASSERT(new_args.size() == 2, "Incorrect number of new arguments");
+    return std::make_shared<FFT>(new_args, inverse, centered);
+}
+//! [op:copy]
+
+//! [op:visit_attributes]
+bool FFT::visit_attributes(ov::AttributeVisitor& visitor) {
+    int inverse_i = static_cast<int>(inverse);
+    int centered_i = static_cast<int>(centered);
+    visitor.on_attribute("inverse", inverse_i);
```
ilyachur, line 172: Why do you need to have static_cast to int?
Visit attributes supports boolean variables from the box.

modules/custom_operations/user_ie_extensions/fft.cpp
ilyachur: PR https://github.com/openvinotoolkit/openvino/pull/14485/ adds parallel_for for OV 2.0 API

modules/custom_operations/user_ie_extensions/fft.cpp
```diff
@@ -0,0 +1,366 @@
...
+        cvScale(out, out, 1.0 / sqrtf(cols), 0);
+
+        cvReleaseMat(&inp);
+        cvReleaseMat(&out);
+    }
+    return true;
+}
+
+bool FFT::has_evaluate() const {
+    return true;
```
ilyachur, line 364: Looks like not all data types are supported.

modules/custom_operations/user_ie_extensions/grid_sample.cpp
```diff
@@ -0,0 +1,124 @@
...
+                    }
+                }
+            }
+        }
+    });
+    return true;
+}
+
+bool GridSample::has_evaluate() const {
+    return true;
```
ilyachur, line 122: The same

modules/custom_operations/user_ie_extensions/sparse_conv_transpose.cpp
ilyachur: Yes this behaviour was changed for the master branch, you can remove empty `visit_attributes` methods.
```suggestion
```

modules/custom_operations/user_ie_extensions/sparse_conv_transpose.cpp
```diff
@@ -0,0 +1,105 @@
...
+                    }
+                }
+            }
+        }
+    }
+    return true;
+}
+
+bool SparseConvTranspose::has_evaluate() const {
+    return true;
```
ilyachur, line 103: The same

ilyachur:
PR https://github.com/openvinotoolkit/openvino/pull/14485 with `ov::parallel_for` was merged to the OpenVINO master branch

modules/custom_operations/user_ie_extensions/fft.cpp
likholat: Thank you, I've updated

modules/custom_operations/user_ie_extensions/ov_extension.cpp
likholat: @slyalin Unfortunately I didn't know about this macro, could you please share some usage examples?

modules/custom_operations/user_ie_extensions/sparse_conv_transpose.cpp
likholat: Thanks, done

modules/custom_operations/README.md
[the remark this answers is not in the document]
likholat: @ilya-lavrenov do you mean to add a flag that specifies which operations should be added?

modules/custom_operations/examples/calculate_grid/export_model.py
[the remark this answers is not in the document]
likholat: @ilyachur, what do you mean? It's not a part of extensions, this script just generates the onnx model. It's needed only for tests and to show how to export torch model with an unsupported operation.

likholat:
https://github.com/openvinotoolkit/openvino_contrib/pull/481#discussion_r1045379180
@ilyachur, could you please clarify what you mean?
This script is not a part of the extension, it's needed only for testing and as an example of operation usage.

ilyachur:
> [#481 (comment)](https://github.com/openvinotoolkit/openvino_contrib/pull/481#discussion_r1045379180) @ilyachur, could you please clarify what you mean? This script is not a part of the extension, it's needed only for testing and as an example of operation usage.

So I mean it is a bug if we cannot load extension by relative path :)

modules/custom_operations/user_ie_extensions/fft.cpp
dkurt: There were issues with boolean attributes (always false) when switched to 2022.1

ilya-lavrenov (approved):

ilya-lavrenov:
With latest changes TBB is not found:
```bash
[ 87%] Building CXX object user_ie_extensions/CMakeFiles/user_ov_extensions.dir/sparse_conv_transpose.cpp.o
In file included from /Users/sandye51/Documents/Programming/git_repo/openvino_contrib/modules/custom_operations/user_ie_extensions/complex_mul.cpp:6:
/Users/sandye51/Documents/Programming/git_repo/openvino/src/core/include/openvino/core/parallel.hpp:42:14: fatal error: 'tbb/blocked_range.h' file not found
#    include "tbb/blocked_range.h"
             ^~~~~~~~~~~~~~~~~~~~~
1 error generated.
make[2]: *** [user_ie_extensions/CMakeFiles/user_ov_extensions.dir/complex_mul.cpp.o] Error 1
In file included from /Users/sandye51/Documents/Programming/git_repo/openvino_contrib/modules/custom_operations/user_ie_extensions/fft.cpp:8:
/Users/sandye51/Documents/Programming/git_repo/openvino/src/core/include/openvino/core/parallel.hpp:42:14: fatal error: 'tbb/blocked_range.h' file not found
#    include "tbb/blocked_range.h"
             ^~~~~~~~~~~~~~~~~~~~~
In file included from /Users/sandye51/Documents/Programming/git_repo/openvino_contrib/modules/custom_operations/user_ie_extensions/grid_sample.cpp:6:
/Users/sandye51/Documents/Programming/git_repo/openvino/src/core/include/openvino/core/parallel.hpp:42:14: fatal error: 'tbb/blocked_range.h' file not found
#    include "tbb/blocked_range.h"
```

ilya-lavrenov:
BTW, we can restore [this workflow](https://github.com/openvinotoolkit/openvino_contrib/actions/runs/3565654445/workflow) instead Azure CI usage.

likholat:
> BTW, we can restore [this workflow](https://github.com/openvinotoolkit/openvino_contrib/actions/runs/3565654445/workflow) instead Azure CI usage.

The idea was to use the OV master branch for testing

.ci/azure/android_arm64.yml
```diff
@@ -122,6 +122,7 @@ jobs:
         -DENABLE_OV_PADDLE_FRONTEND=OFF
         -DBUILD_java_api=ON
         -DBUILD_nvidia_plugin=OFF
+        -DBUILD_custom_operations=OFF
```
ilya-lavrenov, line 125: why do not custom operations work on Android?

ilya-lavrenov (approved):

.ci/azure/linux.yml
```diff
@@ -155,6 +155,23 @@ jobs:
...
 
+  - script: |
+      python3 -m pip install --user virtualenv
+      python3 -m virtualenv -p /usr/bin/python3.8 .env3
+      source .env3/bin/activate
+      python -m pip install --upgrade pip
+      python -m pip install -r $(REPO_DIR)/modules/custom_operations/tests/requirements.txt
+      cd ${OPENVINO_REPO_DIR}/tools && python -m pip install mo/
+    workingDirectory: $(WORK_DIR)
+    displayName: 'Create env'
```
ngaloppo, line 166:
```suggestion
    displayName: 'Create custom ops env'
```

.ci/azure/linux.yml
```diff
@@ -155,6 +155,23 @@ jobs:
...
+
+  - script: |
+      . $(SETUPVARS)
+      source  $(WORK_DIR)/.env3/bin/activate
+      python -m pytest tests/run_tests.py
+    workingDirectory: $(REPO_DIR)/modules/custom_operations
+    displayName: 'Custom Layers tests'
```
ngaloppo, line 173:
```suggestion
    displayName: 'Custom ops tests'
```

README.md
```diff
@@ -11,9 +11,10 @@ This list gives an overview of all modules available inside the contrib reposito
...
 * [**java_api**](./modules/java_api): Inference Engine Java API -- provides Java wrappers for Inference Engine public API.
-* [**mo_pytorch**](./modules/mo_pytorch): PyTorch extensions for Model Optimizer -- native PyTorch to OpenVINO IR converter
+* [**mo_pytorch**](./modules/mo_pytorch): PyTorch extensions for Model Optimizer -- native PyTorch to OpenVINO IR converter.
 * [**optimum**](./modules/optimum): Optimum OpenVINO -- Python extension for Hugging Face [Transformers](https://github.com/huggingface/transformers) including runtime and NNCF quantization.
 * [**Azure Video Analyzer**](./modules/ovms_ai_extension/): Azure Video Analyzer Extension -- enables exchange of video frames and inference results between [Azure Video Analyzer (AVA)](https://docs.microsoft.com/en-us/azure/azure-video-analyzer/video-analyzer-docs/overview) and OpenVINO™ Model Server.
+* [**custom_operations**](./modules/custom_operations/): Collection of Custom Operations -- allows to support Custom Operations with OpenVINO Extensibility Mechanism.
```
ngaloppo, line 17:
```suggestion
* [**custom_operations**](./modules/custom_operations/): Collection of Custom Operations -- implement Custom Operations with OpenVINO Extensibility Mechanism.
```

modules/custom_operations/README.md
```diff
@@ -0,0 +1,61 @@
+# Collection of Custom Operations using OpenVINO Extensibility Mechanism
+
+This module provides a guide and the implementation a few custom operations in the Intel OpenVINO runtime using its [Extensibility Mechanism](https://docs.openvino.ai/latest/openvino_docs_Extensibility_UG_Intro.html).
```
ngaloppo, line 3:
```suggestion
This module provides a guide and the implementation of a few custom operations in the Intel OpenVINO runtime using its [Extensibility Mechanism](https://docs.openvino.ai/latest/openvino_docs_Extensibility_UG_Intro.html).
```

modules/custom_operations/README.md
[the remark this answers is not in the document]
ngaloppo: I agree with @yury-gorbachev: I'm missing some explanation describing what is expected from the pytorch level operation implementation, and how that needs to be propagated to ONNX. It seems to involve implementation of the operation by deriving from `torch.autograd.Function` and overriding the `forward()` and `symbolic()` methods. What is expected from the `symbolic()` function? Do we need to use `ONNX_PASSTHROUGH`, or `ATEN`, or does it not matter? 

Perhaps what we're missing is an example of a pytorch model with only native pytorch ops, and how to create a custom operation for it since it isn't supported by OpenVINO? I thought the FFT example was that, but it seems that it does more than just add a custom operation for `torch.fft`. Should we simplify it?

modules/custom_operations/examples/grid_sample/grid_sample.py
```diff
@@ -0,0 +1,15 @@
...
+import torch.nn.functional as F
+
+class GridSample(torch.autograd.Function):
+    @staticmethod
+    def symbolic(g, x, grid):
+        return g.op('GridSample', x, grid)
+
+    @staticmethod
+    def forward(self, x, grid):
+        return F.grid_sample(x, grid, 'bilinear', 'zeros', True)
```
ngaloppo, line 15: We have native support for GridSample since OpenVINO 2022.2, and ONNX opset 16. Is there still a reason to keep this here, perhaps for legacy reasons?

.ci/azure/android_arm64.yml
ngaloppo: How is OpenVINO runtime's dependency on TBB solved on Android? The custom operations component isn't introducing a new dependency with TBB, is it?

alvoron (changes requested):
If some code is taken from 3rd party repos, you need to add such licenses to third-party-programs.txt hosted in repo root.

dkurt:
> If some code is taken from 3rd party repos, you need to add such licenses to third-party-programs.txt hosted in repo root. Also please contact Yury Semaev to clarify any potential restrictions.

I can confirm that all the code from https://github.com/dkurt/openvino_pytorch_layers (original source of that PR) is written from scratch and signed with Apache 2.0 license:
```
// Copyright (C) 2018-2022 Intel Corporation
```

likholat:
> If some code is taken from 3rd party repos, you need to add such licenses to third-party-programs.txt hosted in repo root. Also please contact Yury Semaev to clarify any potential restrictions.

I've requested a review, waiting for a response

ngaloppo:
As I'm walking through the documentation as a "guide" (in preparation of a blog post), I'm realizing that we're missing the step of how to generate the ONNX model. Right now you need to go and read the test code to understand how everything fits together. I suggest adding a section in the README (e.g. "Example: complex multiplication") in which we describe the steps to convert a PyTorch model with custom operations and execute it with the OpenVINO Runtime:

1. Convert example model to ONNX (`python -m examples.complex_mul.export_model`)
2. Build the complex_mul extension code
3. Run with OpenVINO runtime
  a. Using benchmark_app, after converting to IR with MO
  b. In a Python script, with core.add_extension()

modules/custom_operations/user_ie_extensions/fft.cpp
ilyachur: Is it issue inside the 2022.1 release? I don't think that we need to support old non-LTS release... Master branch should contain the code without WAs which are needed to support old releases

modules/custom_operations/examples/grid_sample/grid_sample.py
likholat: I got an accuracy error when running GridSample tests without the extension

modules/custom_operations/user_ie_extensions/fft.cpp
ngaloppo: @ilya-lavrenov @ilyachur This means we can only test this code in CI with warnings-as-errors turned off. What would be the recommended approach?

ilya-lavrenov: Reimplement cross-platform library loading directory in this library and not rely on OV deprecated API.

ngaloppo: You mean something like this? 

- https://github.com/maddouri/dynalo
- https://github.com/martin-olivier/dylib

Come to think of it, perhaps @dkurt knows why we are loading the OpenCV library at runtime, instead of just linking to it?

dkurt: Because we don't need extra OpenCV dependencies in some projects. That was a requirement to use OpenCV as an optional plugin.

likholat: @dkurt what do you mean by "extra OpenCV dependencies" ?

dkurt: I mean without OpenCV installed nut linked, there will be an error "cannot find libopencv.so". As a plugin we can make this dependency optional.

ngaloppo: I see, in case someone wants to load the extension library on a system without OpenCV, but wants to use any of the operations that do not require OpenCV (like fft), even if fft is compiled into the extension library. 

On the other hand, I see this "contrib" extension library "reference code", not something someone would use immediately in production. If someone really doesn't want to include FFT (with OpenCV dependency), then just don't compile it into the extension library. We even have a way to do that with the `CUSTOM_OPERATIONS` CMake variable (a list of ops to include). 

I am wary of adding a bunch of code (or external dependency) just to show we *can* do this. Instead, I recommend making changes to the build that makes sure the extension library is conditionally linked to OpenCV only when FFT (or any other operation that requires OpenCV) is included in the build.

ngaloppo:
🥳

PR #485, opened 2022-11-29
[ARM plugin] Native support Range operation

allnes:


ilya-lavrenov (approved):

allnes:
@ilya-lavrenov don't merge yet please

allnes:
Input values are as parameters but they need to be as constants, ACL kernel not support current setting

PR #488, opened 2022-11-30
Raise an exception for 3D convolution

alvoron:


modules/arm_plugin/src/arm_converter/arm_converter_conv.cpp
```diff
@@ -206,6 +206,10 @@ struct NEConvolutionLayerQI final: public arm_compute::IFunction {
     std::unique_ptr<arm_compute::NEConvolutionLayer> _conv;
 };
 template<> Converter::Conversion::Ptr Converter::Convert(const opset::ArmConvolution& node) {
+    if (node.get_shape().size() != 4) {
```
allnes, line 209: Failed for shapes less than 4

alvoron: As far as I understand, we may see 2 options: 4 (for 2D) or 5 (for 3D).
There is no option to have less than 4.

PR #524, opened 2022-12-20
[ARM plugin] Reuse common function for query model

allnes:


modules/arm_plugin/src/arm_plugin.cpp
```diff
@@ -96,84 +96,47 @@ InferenceEngine::IExecutableNetworkInternal::Ptr Plugin::LoadExeNetworkImpl(cons
...
     Configuration cfg{config, _cfg, false};
     auto model = network.getFunction();
     if (model == nullptr) {
          IE_THROW() << "Arm Plugin supports only ngraph cnn network representation";
     }
-    std::unordered_set<std::string> originalOps;
-    for (auto&& node : model->get_ops()) {
-        originalOps.emplace(node->get_friendly_name());
-    }
     auto transformedModel = Transform(model, cfg);
```
nkogteva, line 127: Why don't move it inside transformation callback?

nkogteva (approved):

modules/arm_plugin/src/arm_plugin.cpp
allnes: our function `isOperationSupported` is required converter with transformed model that why this transformation callback outside and some tests become is failed

PR #531, opened 2022-12-26
Add transformation that converts layout from NCHW to NHWC

mateusztabaka:
Ticket: 94684


modules/arm_plugin/tests/functional/shared_tests_instances/skip_tests_config.cpp
```diff
@@ -62,5 +62,6 @@ std::vector<std::string> disabledTestPatterns() {
         // Sporadic hanges on linux-debian_9_arm runner (armv7l) 72140
         ".*canStartSeveralAsyncInsideCompletionCallbackWithSafeDtor.*AUTO.*",
 #endif
+        "TransformationTestsF.ConvertLayoutArmMaxPoolV8.*",
```
mateusztabaka, line 65: Tests can be enable after this https://github.com/openvinotoolkit/openvino/pull/14813 is merged.

allnes (approved):
Thank you for PR, graph after transformation looks good. After review @itikhono and @ilya-lavrenov we could merged PR

itikhono:
in general looks good, but I have some doubts about copy-pasting too much code from the core ops.
if we have some changes in the shape inference functions of Convolution, Pool ops (e.g. we found a mistake), we will have to update the arm ops too.
Maybe we can move some checks in common methods/functions (not in this PR). Let's discuss it separetaly.

modules/arm_plugin/src/transformations/convert_layout.cpp
```diff
@@ -0,0 +1,222 @@
...
+    case 5:
+        return std::make_shared<ArmPlugin::opset::Transpose>(input,
+                ArmPlugin::opset::Constant::create(element::i32, Shape{ndhwc_to_ncdhw.size()}, ndhwc_to_ncdhw));
+    default:
+        IE_THROW() << "ConvertLayout: unsupported rank";
+    }
+}
+
+ConvertArmConvolutionLayout::ConvertArmConvolutionLayout() {
+    auto root = ov::pass::pattern::wrap_type<opset::ArmConvolution>(ov::pass::pattern::has_static_rank());
```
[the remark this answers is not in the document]
mateusztabaka: done

mateusztabaka:
@itikhono @allnes please rereview

allnes (approved):
@itikhono please review

modules/arm_plugin/src/opset/conv_arm.hpp
```diff
@@ -21,7 +21,8 @@ class ArmConvolution : public Convolution {
                    const ngraph::CoordinateDiff& pads_begin,
                    const ngraph::CoordinateDiff& pads_end,
                    const ngraph::Strides& dilations,
-                   const ngraph::op::PadType& auto_pad);
+                   const ngraph::op::PadType& auto_pad,
+                   const ngraph::PartialShape& output_shape = {});
```
itikhono, line 25: in what case can this output_shape be empty?

modules/arm_plugin/src/transformations/convert_layout.cpp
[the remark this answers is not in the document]
itikhono: it looks like we're not immune to applying this transformation twice, this could potentially lead to issues.

modules/arm_plugin/src/opset/conv_arm.hpp
mateusztabaka: like here where Convolution is converted to ArmConvolution:  https://github.com/openvinotoolkit/openvino_contrib/blob/c9a60a4d4aa7939595fc451010e5b9b8a545af8b/modules/arm_plugin/src/transformations/conv_bias_fusion.cpp#L158-L163

mateusztabaka:
@allnes there are some issues in azure pipelines, like "unzip: command not found". Is there a way around that?

allnes:
> 

@ilya-lavrenov please help me with answer

allnes:
> @allnes there are some issues in azure pipelines, like "unzip: command not found". Is there a way around that?

@ilya-lavrenov please help me with answer

alvoron:
> > @allnes there are some issues in azure pipelines, like "unzip: command not found". Is there a way around that?
> 
> @ilya-lavrenov please help me with answer, it is problem created last friday and I and Ilya have discussion about it

The issue will be fixed as soon as this PR is merged: https://github.com/openvinotoolkit/openvino_contrib/pull/554

modules/arm_plugin/src/opset/conv_arm.hpp
itikhono: I believe we can replace this transformation with `ConvertArmConvolutionLayout`

itikhono (approved):
I see the CI is green, so we can merge this PR, but in my opinion, we need to take a look at 2 points:

Looks like we can merge ConvertArmConvolutionLayout and ConvertSingleConvolutionToArm functionality and get rid of empty default argument "output_shape = {}" in the operation constructor. After that we will have ArmConvolution always in NHWC layout. 
It's better to avoid the possibility of re-invoking the transformation for the same node in the graph, e.g. 
for some reason we have 2 calls of ConvertArmConvolutionLayout in the transformation pipeline

after 1st call ov::Convolution will be converted to Transpose -> ArmConvolution -> Transpose

after 2nd call Transpose -> ArmConvolution -> Transpose will be converted to Transpose->Transpose->ArmConvolution->Transpose->Transpose

I believe this transformation shouldn't work for ArmConvolution in 2nd call. Maybe we need to try cast Convolution to ArmConvolution inside the transformation and if it's true, just skip it.
Let's do it in the next PR

allnes:
> I see the CI is green, so we can merge this PR, but in my opinion, we need to take a look at 2 points:
> 
> Looks like we can merge ConvertArmConvolutionLayout and ConvertSingleConvolutionToArm functionality and get rid of empty default argument "output_shape = {}" in the operation constructor. After that we will have ArmConvolution always in NHWC layout. It's better to avoid the possibility of re-invoking the transformation for the same node in the graph, e.g. for some reason we have 2 calls of ConvertArmConvolutionLayout in the transformation pipeline
> 
> after 1st call ov::Convolution will be converted to Transpose -> ArmConvolution -> Transpose
> 
> after 2nd call Transpose -> ArmConvolution -> Transpose will be converted to Transpose->Transpose->ArmConvolution->Transpose->Transpose
> 
> I believe this transformation shouldn't work for ArmConvolution in 2nd call. Maybe we need to try cast Convolution to ArmConvolution inside the transformation and if it's true, just skip it. Let's do it in the next PR

@mateusztabaka could you implement this comment? Because when I begin to implement transformations for other layers (this is example for me), I had 2 calls for one transformation

PR #568, opened 2023-01-24
[ARM plugin] delete argument types verification for reference ops

alvoron:


alvoron:
@ilya-lavrenov it seems that deleted exceptions block valid inference sometimes. Without these checks 202-vision-superresolution notebook works fine.
Are you OK to delete these checks? I believe if the plugin doesn't raise an error there, the error will be raised somewhere in nGraph.

ilya-lavrenov (approved):

PR #570, opened 2023-01-25
Rename precisions_array -> precisions_map

mateusztabaka:
Related change: https://github.com/openvinotoolkit/openvino/pull/15277


allnes (approved):
@mateusztabaka please notify when CI wil become green

PR #574, opened 2023-01-27
[ARM CPU plugin] disable TBB

alvoron:


ilya-lavrenov (approved):
Let's port to 2022.3 as well

PR #579, opened 2023-01-30
Update benchmark_app to support NVIDIA Plugin configuration

ytorzuk-altran:
Original benchmark_app doesn't support NVIDIA Plugin configuration. The patch aims to add support of the following NVIDIA Plugin configuration options in the benchmark_app:
* -op_bench option which enables per layer benchmarking to find the most optimal algorithm for each layer
* -nstreams option allows to configure number of CUDA streams

Usage example:
benchmark_app -m /path/to/model -d NVIDIA -ip FP32 -nstreams -op_bench

### DIFF
--- a/modules/nvidia_plugin/utils/openvino_patches/benchmark_app_phase1.patch
+++ b/modules/nvidia_plugin/utils/openvino_patches/benchmark_app_phase1.patch
@@ -1,59 +1,63 @@
-diff --git a/inference-engine/samples/benchmark_app/CMakeLists.txt b/inference-engine/samples/benchmark_app/CMakeLists.txt
-index b37495e5e..9dc5da800 100644
---- a/inference-engine/samples/benchmark_app/CMakeLists.txt
-+++ b/inference-engine/samples/benchmark_app/CMakeLists.txt
-@@ -10,3 +10,4 @@ ie_add_sample(NAME benchmark_app
+diff --git a/samples/cpp/benchmark_app/CMakeLists.txt b/samples/cpp/benchmark_app/CMakeLists.txt
+index 4cb34fe1f4..792f4579c5 100644
+--- a/samples/cpp/benchmark_app/CMakeLists.txt
++++ b/samples/cpp/benchmark_app/CMakeLists.txt
+@@ -10,7 +10,7 @@ file (GLOB HDR ${CMAKE_CURRENT_SOURCE_DIR}/*.hpp)
+ ie_add_sample(NAME ${TARGET_NAME}
+               SOURCES ${SRC}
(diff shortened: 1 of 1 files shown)


modules/nvidia_plugin/utils/openvino_patches/benchmark_app_phase1.patch
```diff
@@ -1,59 +1,63 @@
...
-+add_dependencies(benchmark_app openvino_nvidia_gpu_plugin)
-diff --git a/inference-engine/samples/benchmark_app/benchmark_app.hpp b/inference-engine/samples/benchmark_app/benchmark_app.hpp
-index af18c908e..d55baa34c 100644
---- a/inference-engine/samples/benchmark_app/benchmark_app.hpp
-+++ b/inference-engine/samples/benchmark_app/benchmark_app.hpp
-@@ -62,6 +62,9 @@ static const char enforce_bf16_message[] = "Optional. By default floating point
-                                            "                                  'true'  - enable  bfloat16 regardless of platform support\n"
-                                            "                                  'false' - disable bfloat16 regardless of platform support";
+-              DEPENDENCIES ${GFLAGS_TARGET} format_reader ie_samples_utils)
++              DEPENDENCIES ${GFLAGS_TARGET} format_reader ie_samples_utils openvino_nvidia_gpu_plugin)
```
nkogteva, line 10: Why plugin is required in dependencies?

modules/nvidia_plugin/utils/openvino_patches/benchmark_app_phase1.patch
```diff
@@ -1,59 +1,63 @@
...
- DEFINE_bool(enforcebf16, false, enforce_bf16_message);
+ /// @brief Define flag for showing help message <br>
+ DEFINE_bool(h, false, help_message);
+ 
+@@ -387,6 +390,9 @@ DEFINE_string(scale_values, "", scale_values_message);
+ /// @brief Define flag for inference only mode <br>
+ DEFINE_bool(inference_only, true, inference_only_message);
  
 +/// @brief Optimization mode for particular device
 +DEFINE_bool(op_bench, false, op_bench_mode_message);
```
nkogteva, line 33: Currently we are trying to avoid introduction of device specific properties into the plugin and use `-load_config` option instead of it

nkogteva (commented):
Nvidia plugin have to be updated to support performance hint, because it is generic way to run benchmark_app currently.

nkogteva:
Close since BA supports NVIDIA from the box. `NVIDIA_OPERATION_BENCHMARK` can be set usig config

PR #606, opened 2023-02-24
[Remove APIs] fix issue in arm plugin caused by deprecated run_on_function()

zhaixuejun1993:
Remove call run_on_function(), replaced by run_on_model()

### DIFF
--- a/modules/arm_plugin/src/transformations/finalize_trailing_nodes.hpp
+++ b/modules/arm_plugin/src/transformations/finalize_trailing_nodes.hpp
@@ -9,7 +9,7 @@
 namespace ArmPlugin {
 namespace pass {
 
-class  FinalizeTrailingNodes: public ngraph::pass::FunctionPass {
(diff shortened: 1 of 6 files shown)


modules/arm_plugin/src/transformations/store_result_name.hpp
```diff
@@ -11,7 +11,7 @@ namespace pass {
 class StoreResultName : public ngraph::pass::FunctionPass {
```
ilyachur, line 11:
```suggestion
class StoreResultName : public ov::pass::ModelPass {
```

modules/arm_plugin/src/transformations/finalize_trailing_nodes.hpp
```diff
@@ -12,7 +12,7 @@ namespace pass {
 class  FinalizeTrailingNodes: public ngraph::pass::FunctionPass {
```
ilyachur, line 12: Change to ModelPass

PR #612, opened 2023-03-02
[NVIDIA] Update supported OS in docs

akladiev:
To align with core openvino. Once approved - Jenkins validation will be switched to Ubuntu 20 as well

### DIFF
--- a/modules/nvidia_plugin/README.md
+++ b/modules/nvidia_plugin/README.md
@@ -9,7 +9,7 @@ OpenVINO™ NVIDIA GPU plugin is supported and validated on the following platfo
 
 OS                     | GPU
 ---------------------- | ----------------------
-Ubuntu* 20.04 (64-bit) | Geforce 1080 Ti, NVIDIA T4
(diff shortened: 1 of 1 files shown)


modules/nvidia_plugin/README.md
```diff
@@ -9,7 +9,7 @@ OpenVINO™ NVIDIA GPU plugin is supported and validated on the following platfo
 
 OS                     | GPU
 ---------------------- | ----------------------
-Ubuntu* 18.04 (64-bit) | Geforce 1080 Ti, NVIDIA T4
+Ubuntu* 20.04 (64-bit) | Geforce 1080 Ti, NVIDIA T4
```
nkogteva, line 12: Are list of devices still the same?

akladiev: Quadro RTX 4000 is what we have now

nkogteva: > Quadro RTX 4000 is what we have now

Maybe there is a sense to update it as well?

PR #623, opened 2023-03-14
Finalize SentencePiece tokenizer extensions for MUSE

rkazants:
The following changes are done:
* Create and configure `SentencePieceTokenizer` only once during node construction
* Move to int64 use in the extension
* Use named output ports
* Add documentation


likholat:
Could you please add some information about this operation to the README

I also think, we need an example and tests for this operation.

modules/custom_operations/user_ie_extensions/sentence_piece/sentence_piece.cpp
```diff
@@ -29,80 +29,55 @@ using namespace ov;
...
-SentencepieceTokenizer::SentencepieceTokenizer(const ov::OutputVector& args)
-    : Op(args) {
+SentencepieceTokenizer::SentencepieceTokenizer(const ov::OutputVector& args, int32_t nbest_size, float alpha,
+    bool add_bos, bool add_eos, bool reverse) : m_sp(std::make_shared<SentencePieceProcessor>()),
+    m_nbest_size(nbest_size), m_alpha(alpha), m_add_bos(add_bos), m_add_eos(add_eos),
+    m_reverse(reverse), Op(args) {
+    FRONT_END_GENERAL_CHECK(args.size() == 2, "SentencepieceTokenizer expects two inputs: sp model and input sentences");
+    auto sp_model_const = as_type_ptr<ov::opset10::Constant>(args[0].get_node_shared_ptr());
+    FRONT_END_GENERAL_CHECK(sp_model_const, "SentencepieceTokenizer expects SentencePiece model to be constant.");
+    //sp_model_const->get_data_ptr()
```
rkazants, line 49:
```suggestion
```

dtrawins:
Can we include exception handling for the situating the input data structure is incorrect? Then the input data is random or incorrectly formatted, the extension is causing a segmentation fault. That might be the case then somebody will use benchmark app without -i param or in OVMS when the data is sent as a vector instead of a string.

rkazants:
> Can we include exception handling for the situating the input data structure is incorrect? Then the input data is random or incorrectly formatted, the extension is causing a segmentation fault. That might be the case then somebody will use benchmark app without -i param or in OVMS when the data is sent as a vector instead of a string.

Done

rkazants:
> Could you please add some information about this operation to the README
> 
> I also think, we need an example and tests for this operation.

Thanks for the comment. I will add documentation but there is a problem with programming of such sub-graph with SentencePieceTokenizer for the test. Not easy to do it because I need somewhere to get `sp_model` to initialize the tokenizer.
Propose to proceed without test for it and think of it in the next PR. We test it on the MUSE model manually.

Best regards,
Roman

rkazants:
> Could you please add some information about this operation to the README
> 
> I also think, we need an example and tests for this operation.

Done

ilya-lavrenov (approved):

PR #658, opened 2023-05-19
[NVIDIA] Add IsCudaGraphCompatible() to all operations

apavliuk-altran:
### Details:
- *Add virtual `IsCudaGraphCompatible()` function to `IOperationExec` interface, default `false` value for it in `OperationBase` class and actual values (`true` for the vast majorty) for all inherited operations.*
### Ticket:
- CVS-108535

### DIFF
--- a/modules/nvidia_plugin/src/cuda_operation_base.hpp
+++ b/modules/nvidia_plugin/src/cuda_operation_base.hpp
@@ -42,6 +42,7 @@ public:
                          Inputs inputTensors,
                          Outputs outputTensors,
                          const Workbuffers& workbuffers) const = 0;
+    virtual bool IsCudaGraphCompatible() const = 0;
     virtual void InitSharedImmutableWorkbuffers(const Buffers&) = 0;
     virtual WorkbufferRequest GetWorkBufferRequest() const = 0;
     virtual const WorkbufferIds& GetWorkbufferIds() const = 0;
@@ -75,6 +76,8 @@ public:
                   const ov::Node& node,
                   IndexCollection&& inputIds,
                   IndexCollection&& outputIds);
+
+    bool IsCudaGraphCompatible() const override { return false; }
 
     WorkbufferRequest GetWorkBufferRequest() const override {
         return {};  // Most operators do not need workbuffers
--- a/modules/nvidia_plugin/src/ops/activation_forward_cudnn_base.cpp
+++ b/modules/nvidia_plugin/src/ops/activation_forward_cudnn_base.cpp
@@ -59,5 +59,7 @@ void ActivationForwardCuDnnOpBase::Execute(const InferenceRequestContext& contex
                                                              outputTensors[0].get());
 }
 
+bool ActivationForwardCuDnnOpBase::IsCudaGraphCompatible() const { return true; }
+
 }  // namespace nvidia_gpu
 }  // namespace ov
--- a/modules/nvidia_plugin/src/ops/activation_forward_cudnn_base.hpp
+++ b/modules/nvidia_plugin/src/ops/activation_forward_cudnn_base.hpp
@@ -25,10 +25,13 @@ public:
                                  const ov::Node& node,
                                  IndexCollection&& inputIds,
                                  IndexCollection&& outputIds);
+
     void Execute(const InferenceRequestContext& context,
                  Inputs inputTensors,
                  Outputs outputTensors,
                  const Workbuffers&) const override;
+
+    bool IsCudaGraphCompatible() const override;
 
 protected:
     std::unique_ptr<CUDA::DnnActivationDescriptor> op_desc_;
--- a/modules/nvidia_plugin/src/ops/avgpool.cpp
+++ b/modules/nvidia_plugin/src/ops/avgpool.cpp
@@ -30,6 +30,8 @@ void AvgPoolOp::Execute(const InferenceRequestContext& context,
                   outputs[PoolingImpl::output_index].get());
 }
 
+bool AvgPoolOp::IsCudaGraphCompatible() const { return true; }
+
 OPERATION_REGISTER(AvgPoolOp, AvgPool);
 
 }  // namespace nvidia_gpu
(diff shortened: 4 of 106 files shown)


modules/nvidia_plugin/src/cuda_operation_base.hpp
```diff
@@ -75,6 +76,8 @@ class OperationBase : public IOperationExec, public IOperationMeta, public std::
                   IndexCollection&& inputIds,
                   IndexCollection&& outputIds);
 
+    bool IsCudaGraphCompatible() const override { return false; }
```
nkogteva, line 79: You have 6 operations where this value is false and 47 where it is true. Does it really make sense to modify > 100 files to return the same value?

apavliuk-altran: We made it this way on purpose, so the default value from the base class is false and all new operations wouldn't support `CudaGraph` by default.
So when a new operation is added, the value would be changed to `true` only when a developer made sure about the `CudaGraph` support of this operation.

nkogteva: > We made it this way on purpose, so the default value from the base class is `false` and all new operations wouldn't support `CudaGraph` by default. So when a new operation is added, the value would be changed to `true` only when a developer made sure about the `CudaGraph` support of this operation.

Ok, can we move it into constructor of `OperationBase` with default value `false`?

ytorzuk-altran: [I.25](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#i25-prefer-empty-abstract-classes-as-interfaces-to-class-hierarchies) rule suggests to prefer empty abstract classes as interfaces to class hierarchies since abstract classes that are empty (have no non-static member data) are more likely to be stable than base classes with state.

nkogteva: > [I.25](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#i25-prefer-empty-abstract-classes-as-interfaces-to-class-hierarchies) rule suggests to prefer empty abstract classes as interfaces to class hierarchies since abstract classes that are empty (have no non-static member data) are more likely to be stable than base classes with state.

I didn't suggest modify abstract classes, you can keep inferface empty. I'm talking about `OperationBase` implementation.

nkogteva:
Please add tests

modules/nvidia_plugin/src/cuda_operation_base.hpp
apavliuk-altran: @nkogteva 
If we add a member value to `OperationBase` and initialize it in the constructors we would need approximately the same amount of changes if the vast majority of operations sets `true` while the default value is `false`.
So the number of changed files would be similar.
But if you think this approach is better, we may do it.

apavliuk-altran:
> Please add tests

We couldn't add tests which would test `CudaGraph` feature yet because we don't have of the necessary PR merged yet.
Would it be OK if we add the tests later with another PR?

modules/nvidia_plugin/src/cuda_operation_base.hpp
nkogteva: > @nkogteva If we add a member value to `OperationBase` and initialize it in the constructors we would need approximately the same amount of changes if the vast majority of operations sets `true` while the default value is `false`. So the number of changed files would be similar. But if you think this approach is better, we may do it.

2x less files will be touched. But it is up yo you, for me it looks like overkill.

nkogteva:
> > Please add tests
> 
> We couldn't add tests which would test `CudaGraph` feature yet because we don't have of the necessary PR merged yet. Would it be OK if we add the tests later with another PR?

I'm talking about tests which will check expected value from `IsCudaGraphCompatible` per operation.

nkogteva:
> > > > Please add tests
> > > 
> > > 
> > > We couldn't add tests which would test `CudaGraph` feature yet because we don't have of the necessary PR merged yet. Would it be OK if we add the tests later with another PR?
> > 
> > 
> > I'm talking about tests which will check expected value from `IsCudaGraphCompatible` per operation.
> 
> The thing is we faced the fact that graph compatibility may depend on hardware or/and environment. We already have faced such behavior with `LSTMSequence` operation and now we try to determine what specifically affects it. So, from our side, we propose to add rather functional tests for the complete CUDA Graph feature than hardcode the tests for some operations. This way, we may quickly find if something goes wrong with another version of a library or hardware.

Taking it into account I should ask does it make sense to split feature to small PRs? It is hard to add tests and probably functionality will be changed in future.

apavliuk-altran:
> > > > > Please add tests
> > > > 
> > > > 
> > > > We couldn't add tests which would test `CudaGraph` feature yet because we don't have of the necessary PR merged yet. Would it be OK if we add the tests later with another PR?
> > > 
> > > 
> > > I'm talking about tests which will check expected value from `IsCudaGraphCompatible` per operation.
> > 
> > 
> > The thing is we faced the fact that graph compatibility may depend on hardware or/and environment. We already have faced such behavior with `LSTMSequence` operation and now we try to determine what specifically affects it. So, from our side, we propose to add rather functional tests for the complete CUDA Graph feature than hardcode the tests for some operations. This way, we may quickly find if something goes wrong with another version of a library or hardware.
> 
> Taking it into account I should ask does it make sense to split feature to small PRs? It is hard to add tests and probably functionality will be changed in future.

With smaller PRs it is easier to perform several tasks simultaneously for more than one person. Also, they are easier to review.

As https://github.com/openvinotoolkit/openvino_contrib/pull/657 was already merged, we will add some tests using Cuda Graph functionality and `IsCudaGraphCompatible` fucntion.
Also, we still have some changes pending to this PR, so we'll update it soon.

nkogteva (approved):
In general, LGTM

nkogteva:
Please merge https://github.com/apavliuk-altran/openvino_contrib/pull/1/

PR #699, opened 2023-08-04
Fix OpenCV path for non-Linux platforms

rajatkrishna:
## Details

Currently, the build scripts for the Java sample applications resolve the OpenCV jar location for Linux installations correctly based on the `OpenCV_DIR` env variable but fail on Windows platforms. 
This PR updates the build scripts to locate the jar based on the OS.


likholat:
@ilya-lavrenov we can close this PR since we already have these changes for master branch

PR #701, opened 2023-08-08
Rename builder namespace1

olpipi:


olpipi:
Is blocked by https://github.com/openvinotoolkit/openvino/pull/19003 and https://github.com/openvinotoolkit/openvino/pull/19026

PR #702, opened 2023-08-08
[NVIDIA] Set proper math type for convolution

apavliuk-altran:
### Details:
- *Sets proper convolution math type according to the math type of the selected algo*

### DIFF
--- a/modules/nvidia_plugin/src/ops/convolution_components/convolution_cudnn_components.cpp
+++ b/modules/nvidia_plugin/src/ops/convolution_components/convolution_cudnn_components.cpp
@@ -74,7 +74,7 @@ CUDA::DnnConvolutionDescriptor ConvolutionParamsCuDnn::MakeConvolutionDescriptor
 
     // Enable computations on Tensor Core hardware which requires at least Volta GPU
     // (compute capability 7.0).
-    const cudnnMathType_t math_type = CUDNN_TENSOR_OP_MATH_ALLOW_CONVERSION;
+    const cudnnMathType_t math_type = CUDNN_TENSOR_OP_MATH;
     throwIfError(::cudnnSetConvolutionMathType(conv_desc.get(), math_type));
     throwIfError(::cudnnSetConvolutionGroupCount(conv_desc.get(), groups_));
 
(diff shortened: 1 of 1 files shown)


nkogteva:
And what about hardcoded value `CUDNN_TENSOR_OP_MATH_ALLOW_CONVERSION`, which you try to change previosly? Should we avoid convertion if user specify inference precision == `f32`?

apavliuk-altran:
> And what about hardcoded value `CUDNN_TENSOR_OP_MATH_ALLOW_CONVERSION`, which you try to change previosly? Should we avoid convertion if user specify inference precision == `f32`?

We tried `CUDNN_TENSOR_OP_MATH` value but it didn't help to solve the precision issue for some tests with FP32 precision.
So we left the previous value, it won't affect FP16 precision.
For FP32 precision, I think, we should leave it as is because there were no problems with FP32 accuracy with this setting.

nkogteva:
> > And what about hardcoded value `CUDNN_TENSOR_OP_MATH_ALLOW_CONVERSION`, which you try to change previosly? Should we avoid convertion if user specify inference precision == `f32`?
> 
> We tried `CUDNN_TENSOR_OP_MATH` value but it didn't help to solve the precision issue for some tests with FP32 precision. So we left the previous value, it won't affect FP16 precision. For FP32 precision, I think, we should leave it as is because there were no problems with FP32 accuracy with this setting.

It's not just about tests. If the user has explicitly specified an inference precision f32, we should avoid converting to lower precision, is not it?

apavliuk-altran:
> > > And what about hardcoded value `CUDNN_TENSOR_OP_MATH_ALLOW_CONVERSION`, which you try to change previosly? Should we avoid convertion if user specify inference precision == `f32`?
> > 
> > 
> > We tried `CUDNN_TENSOR_OP_MATH` value but it didn't help to solve the precision issue for some tests with FP32 precision. So we left the previous value, it won't affect FP16 precision. For FP32 precision, I think, we should leave it as is because there were no problems with FP32 accuracy with this setting.
> 
> It's not just about tests. If the user has explicitly specified an inference precision f32, we should avoid converting to lower precision, is not it?

Changed to `CUDNN_TENSOR_OP_MATH`

nkogteva (approved):

PR #713, opened 2023-08-29
OpenVINO Code VSCode Extension

apaniukov:


alvoron (changes requested):
1. Root repository `README` is not modified - new module should be mentioned there.
2. `LICENSE` file is empty - do we really need dedicated LICENSE file? May we use root LICENSE?
3. Third party dependencies (if any) should be added to `third-party-programs.txt`
4. Any new Azure pipelines to validate the module?

yatarkan:
> 1. Root repository `README` is not modified - new module should be mentioned there.
> 2. `LICENSE` file is empty - do we really need dedicated LICENSE file? May we use root LICENSE?
> 3. Third party dependencies (if any) should be added to `third-party-programs.txt`
> 4. Any new Azure pipelines to validate the module?

@alvoron Thanks for your review.
1. I  have updated root `README.md` file with short description of new module
2. Empty license file was deleted - let's use root license for now
3. If I understand it correctly, we do not have anything to add to the `third-party-programs.txt`. We do have some dependencies (e.g. npm packages and pip modules installed from package managers), however I assume all of them should not be listed in `third-party-programs.txt` as we do not include them as git submodules or "as-is" source code files and we do not redistribute them. Please correct me if I am missing something here.
4. We don't have automatic validation through CI for now, but we are planning to add it later.

Should we also make changes in `.github/CODEOWNERS` file and add separate maintainers team (e.g. `@openvinotoolkit/openvino_contrib-openvino_code-maintainers`) for newly added module?

yatarkan (approved):

andrei-kochin (approved):

ilya-lavrenov (approved):

PR #732, opened 2023-09-25
[NVIDIA] DetectionOutput fix input types

apavliuk-altran:
### Details:
- *DetectionOutput operation in NVIDIA plugin is missing a check to assure that all the input/output types are the same.
In case all types are FP16 but the 3rd input type is FP32 the operation result is incorrect.
This PR adds the proper check and a transformation to covert any inputs with incorrect type*
### Ticket:
- 119127
### Note:
- *According to DetectionOutput specification (https://docs.openvino.ai/2023.0/openvino_docs_ops_detection_DetectionOutput_1.html):
"The first tuple with batch_id equal to -1 means end of output."
So the implementation in NVIDIA plugin doesn't change the rest of the values in the output after the 1st '-1.0' value and all other values should be ignored.*

### DIFF
--- a/modules/nvidia_plugin/src/transformer/detection_output_fix_input_types_transformation.cpp
+++ b/modules/nvidia_plugin/src/transformer/detection_output_fix_input_types_transformation.cpp
@@ -1,0 +1,57 @@
+// Copyright (C) 2018-2023 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#include "detection_output_fix_input_types_transformation.hpp"
+
+#include "openvino/cc/pass/itt.hpp"
+#include "openvino/core/rt_info.hpp"
+#include "openvino/op/convert.hpp"
+#include "openvino/op/detection_output.hpp"
+#include "openvino/pass/pattern/op/pattern.hpp"
+#include "openvino/pass/pattern/op/wrap_type.hpp"
+
+using namespace ov::pass::pattern;
+
+namespace ov::nvidia_gpu::pass {
+
+bool detection_output_fix_input_types(Matcher& m) {
+    const auto d_out = std::dynamic_pointer_cast<ov::op::v0::DetectionOutput>(m.get_match_root());
+    if (!d_out) {
+        return false;
+    }
+    const auto& type = d_out->get_element_type();
+    ov::OutputVector inputs;
+    bool needs_transform = false;
+    for (std::size_t i = 0; i < d_out->inputs().size(); ++i) {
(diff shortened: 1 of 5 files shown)


modules/nvidia_plugin/src/transformer/detection_output_fix_input_types_transformation.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+    const auto m = std::make_shared<Matcher>(d_out, matcher_name);
+    register_matcher(m, callback);
+}
+
+DetectionOutputFixInputTypesTransformation5::DetectionOutputFixInputTypesTransformation5() {
+    MATCHER_SCOPE(DetectionOutputFixInputTypesTransformation5);
+
+    const auto d_out =
+        wrap_type<ov::op::v0::DetectionOutput>({any_input(), any_input(), any_input(), any_input(), any_input()});
```
nkogteva, line 77: Why do we need 2 similiar matchers? Why not use just `wrap_type<ov::op::v0::DetectionOutput>();`?

modules/nvidia_plugin/src/transformer/detection_output_fix_input_types_transformation.cpp
```diff
@@ -0,0 +1,84 @@
...
+    if (!needs_transform) {
+        return false;
+    }
+
+    const auto inputs_size = input_nodes.size();
+    const auto& attrs = d_out->get_attrs();
+    std::shared_ptr<ov::op::v0::DetectionOutput> new_d_out{nullptr};
+    if (inputs_size == 3) {
+        new_d_out =
+            std::make_shared<ov::op::v0::DetectionOutput>(input_nodes[0], input_nodes[1], input_nodes[2], attrs);
```
nkogteva, line 49: Can we use `clone_with_new_inputs` here?

nkogteva (changes requested):
Please add transformation tests

modules/nvidia_plugin/src/transformer/detection_output_fix_input_types_transformation.cpp
apavliuk-altran: Done

modules/nvidia_plugin/src/transformer/detection_output_fix_input_types_transformation.cpp
apavliuk-altran: Done

apavliuk-altran:
> Please add transformation tests

Done

nkogteva (approved):

PR #745, opened 2023-10-10
[NVIDIA] Fix Unit Tests

redradist:
## Description

Changes:
- Previously in release mode some of Unit Tests expected to crash with ASSERT_DEATH, but after switching to OPENVINO_ASSERT which always throws exception it is not anymore a case
- Updated MatMul test which not always converts to FP16


nkogteva (approved):

redradist:
@nkogteva I suggest to enable Unit Tests for NVIDIA plugin by default, to avoid such thing in future

PR #754, opened 2023-11-02
Add references to Rust API

abrown:
This change adds documentation, but no code, to explain how one might use OpenVINO from Rust. This is a partial response to https://github.com/intel/openvino-rs/issues/70; there are more things that need to be done before we re-evaluate moving any code.


abrown:
cc: @redradist, @ilya-lavrenov, @alvoron

ilya-lavrenov (approved):

PR #757, opened 2023-11-09
[VSCode Extension] Enable code generation with user selected text only (CVS-119131)

luke-lin-vmc:
### Enable code generation with user selected text only (CVS-119131)

For inline code completion context is used and it is sent to the generation model as input.
Now this context/input is a text from the beginning of the file till current cursor position.

The idea is to add mechanism that will limit this context by sending to generation model only text explicitly selected by user.

It is an open question how to implement it from the UI/UX perspective, how it should look like and how user should interact with it. 
Such feature and interaction examples can be found in other similar VSCode extensions (e.g. refact.ai, cody, etc.)

### DIFF
--- a/modules/openvino_code/src/inline-completion/completion.service.ts
+++ b/modules/openvino_code/src/inline-completion/completion.service.ts
@@ -31,6 +31,15 @@ class CompletionService {
     if (fillInTheMiddleMode && textAfterCursor.trim()) {
       return `${startToken}${textBeforeCursor}${middleToken}${textAfterCursor}${endToken}`;
     }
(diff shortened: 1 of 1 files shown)


apaniukov:
@luke-lin-vmc thanks for the contribution.
Could you update README.md and the side panel description of the code completion feature?

andrei-kochin:
@luke-lin-vmc please address lint issues in failed CI

luke-lin-vmc:
Error happens if "--max-warnings 0" is set. Will modify the code then submit new PR

PR #772, opened 2023-12-06
Fix Windows warnings

Wovchena:


ilya-lavrenov (approved):
but we will migrate to string tensors soon

PR #786, opened 2023-12-13
[VSCode Extension] UI enhancements

luke-lin-vmc:
1. Generate code with user selected text only
2. Add right-click menus, "Generate Inline Code Completion" and "Generate Inline Code Completion In New Tab"


apaniukov (approved):

akladiev:
build_jenkins

PR #807, opened 2023-12-27
[NVIDIA] Move cuda graphs instantiation to model compilation

ytorzuk-altran:
Details:

- *Cuda graphs instantiation was moved to model compilation*
Ticket:
- CVS-119834


p-durandin:
build_jenkins

modules/nvidia_plugin/src/cuda_compiled_model.cpp
```diff
@@ -278,6 +284,65 @@ std::shared_ptr<ov::IAsyncInferRequest> CompiledModel::create_benchmark_infer_re
         get_callback_executor());
 }
 
+#include <condition_variable>
```
redradist, line 287: Please, move `#include`-s at top of file

PR #809, opened 2023-12-28
Delete API1.0 behavior/ tests

vurusovs:
In order to remove IE/ngraph API1.0 need to delete old behavior tests


vurusovs:
build_jenkins

vurusovs:
build_jenkins

PR #823, opened 2024-01-14
[custom operations] Use ov::element in custom operations module

praasz:
Use `ov::element` instead of 'ngraph::element` to fix build issues after removing `HostTensor` in openvinotoolkit/openvino#22146


praasz:
build_jenkins

ilya-lavrenov (approved):

PR #848, opened 2024-01-25
[OPENVINO CODE] Added Fill-in-the-middle(FIM) support 

kumarijy:
Adding FIM support through deepseek-coder-1.3b-instruct and changed model names at Frontend

### DIFF
--- a/modules/openvino_code/package.json
+++ b/modules/openvino_code/package.json
@@ -193,7 +193,7 @@
             "default": "code-t5",
             "enum": [
               "code-t5",
-              "decicoder-1b-openvino-int8",
+              "decicoder-1b-openvino",
               "stablecode-completion",
               "deepseek-coder"
             ],
--- a/modules/openvino_code/package.json
+++ b/modules/openvino_code/package.json
@@ -1,7 +1,7 @@
 {
   "publisher": "OpenVINO",
   "name": "openvino-code-completion",
-  "version": "0.0.6",
+  "version": "0.0.8",
   "displayName": "OpenVINO Code Completion",
   "description": "VSCode extension for AI code completion with OpenVINO",
   "icon": "media/logo.png",
--- a/modules/openvino_code/package.json
+++ b/modules/openvino_code/package.json
@@ -190,11 +190,12 @@
           "openvinoCode.model": {
             "order": 0,
             "type": "string",
-            "default": "codet5p-220m-py",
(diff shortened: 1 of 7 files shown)


modules/openvino_code/package.json
```diff
@@ -207,6 +208,40 @@
...
+            "order": 4,
+            "type": "boolean",
+            "default": "false",
+            "description":
+              "When checked, text before (above) and after (below) the cursor will be used for completion generation. When unckecked, only text before (above) the cursor will be used."
+          },
+          "openvinoCode.startToken": {
+            "order": 7,
+            "type": "string",
+            "default": "<fim_prefix>",
```
apaniukov, line 221: `deepseek-coder` has a different FIM tokens format:  `"<｜fim▁begin｜>"`, check [tokenizer.json](https://huggingface.co/deepseek-ai/deepseek-coder-1.3b-instruct/raw/main/tokenizer.json)

modules/openvino_code/shared/model.ts
```diff
@@ -4,22 +4,27 @@ enum ModelId {
   CODE_T5_220M = 'Salesforce/codet5p-220m-py',
   DECICODER_1B_OPENVINO_INT8 = 'chgk13/decicoder-1b-openvino-int8',
   STABLECODE_COMPLETION_ALPHA_3B_4K_OPENVINO_INT8 = 'chgk13/stablecode-completion-alpha-3b-4k-openvino-int8',
+  //deepseek_coder = 'deepseek-ai/deepseek-coder-6.7b-instruct',
+  deepseek_coder_7b = 'kumarijy/deepseek-coder-1.3b-instruct_ov',
```
apaniukov, line 8: Use ALL_CAPS for consistency. It is also a 1_3B model, not a 7B.

modules/openvino_code/package.json
apaniukov: I think it makes more sense to make a FIM feature a check box and to hardcode FIM tokens, as long as the model list is predefined.

apaniukov (changes requested):
You must include a licence file to the HF model repository based on [this](https://huggingface.co/deepseek-ai/deepseek-coder-1.3b-instruct/blob/0f12b153ed35650259fa19984891872d81a61eda/LICENSE#L46).

apaniukov (approved):

PR #854, opened 2024-01-26
[Tokenizers][TF FE] Fix MUSE conversion

apaniukov:


rkazants (commented):
please productize changes and let us merge it

modules/custom_operations/user_ie_extensions/tokenizer/sentence_piece.cpp
```diff
@@ -157,24 +134,14 @@ bool SentencepieceTokenizer::evaluate(TensorVector& outputs, const TensorVector&
     if(input_element_type == ov::element::string) {
```
rkazants, line 134: Please remove artifacts of packed strings: `begin_ids`, etc. above

rkazants (approved):
one minor comment remained

modules/custom_operations/user_ie_extensions/tokenizer/sentence_piece.cpp
apaniukov: Deleted.

modules/custom_operations/user_ie_extensions/tokenizer/tensorflow_translators.cpp
```diff
@@ -197,18 +178,9 @@ ov::OutputVector translate_reshape(const ov::frontend::NodeContext& node) {
...
-        auto reshape = post_translate_string_tensor_output({begins, ends, chars});
-        return {reshape};
-    } else {
-        auto reshape = std::make_shared<Reshape>(tensor, shape, false);
-        return {reshape};
-    }
+    auto reshape = std::make_shared<Reshape>(tensor, shape, false);
+    return {reshape};
+//    }
     // set_node_name(node.get_name(), reshape); // TODO: requires dependencies from TF FE internals
```
rkazants, line 184: do we need this? Please clean the code a bit.

rkazants, line 184: I think you should revert these changes, they does not relate MUSE model

apaniukov: Deleted

apaniukov: Reverted.

rkazants (approved):

apaniukov:
@dtrawins @slyalin sentencepiece operation translated from TF with string input by default. It also allows input of a decomposed string representation, but this must be changed manually after the model is converted. This PR allows you to experiment with data format optimizations without having to recompile the extension.

PR #861, opened 2024-01-31
[TOKENIZERS] Set module version

mryzhov:


ilya-lavrenov (approved):

apaniukov:
Could we add a version of openvino, that is compatible with tokenizers?

apaniukov (approved):

PR #888, opened 2024-03-06
Do not use opset in transformation tests

olpipi:


olpipi:
Prior fix to https://github.com/openvinotoolkit/openvino/pull/22725

olpipi:
build_jenkins

ilya-lavrenov (approved):

PR #919, opened 2024-05-27
[CI] Add `overall_status` jobs to GHA workflows

akashchi:


.github/workflows/linux.yml
```diff
@@ -2,16 +2,10 @@ name: Linux (Ubuntu 20.04, Python 3.11)
 on:
   workflow_dispatch:
   pull_request:
-    paths-ignore:
-        - 'modules/nvidia_plugin'
-        - 'modules/openvino_code'
```
akashchi, line 7: It is not needed as the NVIDIA plugin job was integrated into this workflow.

.github/workflows/linux.yml
```diff
@@ -331,3 +325,17 @@ jobs:
 
       - name: Show ccache stats
         run: ccache --show-stats
+
+  Overall_Status:
```
akashchi, line 329: It is not required as of now (as I've removed the conditional start of this workflow) yet it could be beneficial in the future, for example for Smart CI enabling.

akladiev (approved):

PR #940, opened 2024-11-27
CVS-157213-fixing vulnerabilities - [NPM] Fix npm audit findings in ov.contrib 

kumarijy:
CVS-157213-fixing vulnerabilities - [NPM] Fix npm audit findings in ov.contrib 
Ran npm audit fix --force to remove vulnerabilities
Also, updated OV version to 2024.5.0

### DIFF
--- a/modules/openvino_code/package.json
+++ b/modules/openvino_code/package.json
@@ -1,7 +1,7 @@
 {
   "publisher": "OpenVINO",
   "name": "openvino-code-completion",
-  "version": "0.1.0",
+  "version": "0.2.0",
   "displayName": "OpenVINO Code Completion",
   "description": "VSCode extension for AI code completion with OpenVINO",
   "icon": "media/logo.png",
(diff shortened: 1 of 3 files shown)


modules/openvino_code/package.json
```diff
@@ -84,7 +83,7 @@
     "terser-webpack-plugin": "5.3.9",
     "ts-loader": "9.4.4",
     "typescript": "5.1.6",
-    "webpack": "5.88.2",
+    "webpack": "^5.96.1",
```
gblong1, line 86: Just double confirming that this ^ character is valid

gblong1 (changes requested):
Added one comment, otherwise LGTM.

modules/openvino_code/package.json
kumarijy: yes this ^ character is valid. It just shows that the webpack package has been updated automatically once you run npm install

apaniukov (approved):

PR #969, opened 2025-05-20
OpenVINO Code Completion is no longer maintained.

gblong1:
Removing openvino_code project from openvino_contrib. This project will no longer be maintained and code is removed to avoid any outstanding issues with dependabot vulnerability checking.


kumarijy (approved):
The changes looks good to me

PR #977, opened 2025-07-01
[langchainJS] Fix embeddings

Retribution98:
## Details:
- Fixed `OpenVINOEmbeddings` work with the embedding models.
- Expanded list of tested embedding models

### DIFF
--- a/modules/openvino-langchain/sample/README.md
+++ b/modules/openvino-langchain/sample/README.md
@@ -19,5 +19,3 @@ cd sample/
 npm install
 node index.js *path_to_llm_model_dir* *path_to_embeddings_model_dir*
 ```
-
-The sample model `Phi-3.5-mini-instruct` should include both LLM and embeddings models.


modules/openvino-langchain/src/embeddings.ts
```diff
@@ -100,30 +103,25 @@ export class OpenVINOEmbeddings extends Embeddings {
...
+    const tokenizerOutputs = tokenizerModelCompiled.outputs
+      .map(o => o.getAnyName())
+      .sort();
+    const embeddingInputs = modelCompiled.inputs
+      .map(i => i.getAnyName())
+      .sort();
+    if (!tokenizerOutputs.every(
+      (value, index) => value === embeddingInputs[index],
+    )) {
+      throw Error('Embedding model is incorrect.');
```
almilosz, line 118:
```suggestion
      throw Error('Embedding model is incorrect. Tokenizer outputs should be the same as embedding model inputs.');
```

modules/openvino-langchain/README.md
```diff
@@ -21,11 +21,17 @@ npm install openvino-langchain
 
 In order to use OpenVINO, you need to convert and compress the text generation model into the [OpenVINO IR format](https://docs.openvino.ai/2025/documentation/openvino-ir-format.html).
 
-Tested compatibility with:
-  - BAAI/bge-small-en-v1.5 (Embeddings model)
-  - openlm-research/open_llama_7b_v2
-  - meta-llama/Llama-2-13b-chat-hf
-  - microsoft/Phi-3.5-mini-instruct
+The following models are tested:
```
almilosz, line 24: In the  _openvino-langchain/sample/README.md_ there is a mention that `Phi-3.5-mini-instruct` includes embedding model. It also should be updated.

almilosz (approved):
Good catch! This PR shows the proper use of `openvino-node` for embeddings creation.

I tested it with [Llama-3.1-8B-Instruct-FastDraft-150M](https://huggingface.co/OpenVINO/Llama-3.1-8B-Instruct-FastDraft-150M-int8-ov) and [bge-small-en-v1.5](BAAI/bge-small-en-v1.5)

modules/openvino-langchain/README.md
Retribution98: Thanks, this row is removed!

PR #1002, opened 2025-09-24
Add GenAI Optimizations module

l-bat:
### Changes

Integration of [CDPruner](https://arxiv.org/pdf/2506.10967) algorithm to new GenAI Optimizations Module in OpenVINO Contrib. The goal of this module is to provide a PoC for various optimization methods in GenAI pipelines before productization into OpenVINO GenAI.

### Reason for changes

This integration removes the dependency on the outdated fixed Transformers version from the POC repository and simplifies experimentation with new methods..

### Related tickets

169957


l-bat:
@alvoron FYI about the new module

ljaljushkin (approved):

PR #1012, opened 2025-10-22
[GenAI] Support Token Eviction for LRMs

l-bat:
## Changes

* Integrated the Adaptive R-KV Token Eviction algorithm, which computes token importance within a small sliding window of the most recent queries during both the prefill and decode stages.
* Added refined selection methods (KVCrush and DiverseKV) to improve token diversity and reduce redundancy
* Added reasoning benchmark (MATH500 and GSM8K datasets)
Integrated the Adaptive R-KV Token Eviction algorithm, which computes token importance within a small sliding window of the most recent queries during both the prefill and decode stages.

## Related tickets
169957


l-bat:
@openvinotoolkit/openvino_contrib-maintainers please take a look at the `custom_operations` module tests — it seems that the CI is broken.

PR #1021, opened 2025-11-18
Do not use deprecated ov::Model ctr

olpipi:


olpipi:
Leftovers of https://github.com/openvinotoolkit/openvino/pull/32840

t-jankowski (approved):

PR #1024, opened 2025-12-10
[GenAI] Fix cgo callback pointer handling in genai

zxjzxj9:
- wrap Sequence in cgo.Handle before passing into C callbacks
- unwrap handle in goCallbackBridge and delete after generation
- prevents cgo “Go pointer to Go pointer” panic during OpenVINO runs
- make cgo compatible accross different openvino versions


zxjzxj9:
Hi @zhaohb, is there any update for this MR?

PR #1025, opened 2025-12-13
fix changes in

josejuanmontiel:
C.OV_GENAI_STREAMMING_STATUS


allnes:
@josejuanmontiel hi Is this request still relevant?

josejuanmontiel:
> @josejuanmontiel hi Is this request still relevant?

Relevant? Seem that from 2025.02 to 04 fix this typo en a constant streamming... With double mm...

And when I try to run olllama with open vino libs in this version need to fix this...

For my was a way of try to learn things... Maybe you have a more clear vision about version supported an the roadmap 

Maybe when I open the PR should used the supported version but because I use Debian maybe was different for me... It was hard to install (for me) outside Ubuntu repository...

Sorry no more context for my side

Thanks

PR #1037, opened 2026-02-16
[NVIDIA] Fixed build of NVIDIA plugin

redradist:


modules/nvidia_plugin/src/cuda_plugin.cpp
```diff
@@ -78,15 +78,15 @@ std::shared_ptr<ov::ICompiledModel> Plugin::compile_model(const std::shared_ptr<
 }
 
 std::shared_ptr<ov::ICompiledModel> Plugin::import_model(const ov::Tensor& model, const ov::AnyMap& properties) const {
-    ov::SharedStreamBuffer buffer{model.data(), model.get_byte_size()};
+    ov::SharedStreamBuffer buffer{static_cast<const char*>(model.data()), model.get_byte_size()};
```
p-durandin: Previous code used in OV main repository. Why is it required to change?

praasz: This is strange the current version of OV has `ov::SharedStreamBuffer ` which accepts the `const void*` the cast is not required.

p-durandin: Please provide more details about compiler' version, OS and etc. We'd like to unblock further merges but need to avoid extra casts if not required

redradist: gcc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0

redradist: > This is strange the current version of OV has `ov::SharedStreamBuffer ` which accepts the `const void*` the cast is not required.

Issue was due to flag `-fpermissive`

redradist: Yes, agree .. it was some issue in my environment, mismatched version of OpenVINO and open_contrib. Closing PR

PR #1044, opened 2026-03-11
[JAVA_API] Fix wrapper close bug

ClaudioPaul0:
Corrects the logic in `Wrapper.close()` to ensure that the native resource is released.
Without this fix, no memory is being released when close is called once.

### DIFF
--- a/modules/java_api/src/main/java/org/intel/openvino/Wrapper.java
+++ b/modules/java_api/src/main/java/org/intel/openvino/Wrapper.java
@@ -27,7 +27,7 @@ public class Wrapper implements AutoCloseable {
 
     @Override
     public void close() {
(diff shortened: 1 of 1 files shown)


shivareddi35-a11y (commented):
The current implementation calls compareAndSet twice with conflicting conditions,
which prevents the native resource from ever being released.

The first call already sets the closed flag to true, making the second
compareAndSet(false, true) impossible.

The correct approach is to perform the check only once and release the
native resource when the state changes from false to true.

ivolucas (approved):

allnes (approved):

allnes:
Hi @ClaudioPaul0, could you fix CI please?

PR #1064, opened 2026-05-25
SAM6D Enabled with OpenVino

pravin25:
SAM6D Enabled with OpenVino


pj3iL:
A few suggestions:
- links in the top level readme file are wrong for sam-6d.
- As the repo is for OpenVino, rather than "OpenVino SAM6D" use SAM6D, specially in the top level readme file.
- In modules/3d/OV-SAM-6D/README.md, it will be good if original repo with commit id can be mentioned, and not in modules/3d/OV-SAM-6D/SAM-6D/OV_README.md.
- plyfile license is duplicated in third-party-programs.txt, please check for any other repetition.

p-durandin:
build_jenkins

p-durandin:
build_jenkins

alvoron:
@allnes could you please review it?

modules/3d/OV-SAM-6D/SAM-6D/Instance_Segmentation_Model/run_ism.sh
```diff
@@ -0,0 +1,114 @@
...
+  log "Log file: $run_log"
+
+  # Run inference in background, write output to log file directly
+  # so the log is always written even if stdout is piped through head/grep.
+  local rc=0
+  OV_DEVICE="$OV_DEVICE" \
+  OV_SAM_DEVICE="$OV_SAM_DEVICE" \
+  PYTHONUNBUFFERED=1 \
+  conda run --no-capture-output -n "$CONDA_ENV" \
+    "$PYTHON_BIN" run_inference_ov_10.py dataset_name="$DATASET" \
```
allnes: `run_inference_ov_10.py` does not exist: it is not in this PR, not created by `ism.patch`/`pem.patch`, and not in upstream SAM-6D at `1c2543b`. This script cannot run. The BOP evaluator shipped here is `eval_ism_ov_bop.py`, which uses a different CLI (`--bop_dir`, `--ext`, argparse) rather than the hydra-style `dataset_name=`. Please repoint this at `eval_ism_ov_bop.py` and fix the invocation, or add the missing runner.

modules/3d/OV-SAM-6D/SAM-6D/Pose_Estimation_Model/run_inference_custom_pytorch.py
```diff
@@ -0,0 +1,484 @@
...
+        pred_R, pred_t, pred_pose_score = model(*model_input_tuple)
+        pred_R = pred_R.detach().cpu().numpy()
+        pred_t = pred_t.detach().cpu().numpy()
+        pred_pose_score = pred_pose_score.detach().cpu().numpy()
+        # out = model(input_data)
+    pem_time = time.time() - pem_time_start
+    print(f"[PyTorch] pose estimation inference time: {pem_time*1000:.2f} ms")
+
+    # pytorch save results
+    if 'pred_pose_score' in input_data.keys():
```
allnes: `input_data` never contains the key `'pred_pose_score'`: the model returns it into a local variable at line 453. This condition is always `False`, the `else` at line 462 always runs, and the computed `pred_pose_score` is discarded, so the final pose score never uses the model's confidence.

Suggest: `pose_scores = pred_pose_score * input_data['score'].detach().cpu().numpy()`

modules/3d/OV-SAM-6D/SAM-6D/Pose_Estimation_Model/run_inference_custom_pytorch.py
```diff
@@ -0,0 +1,484 @@
...
+    default_output_dir = os.path.join(project_root, 'SAM-6D/Data/Example/outputs')
+    default_seg_path = os.path.join(default_output_dir, 'sam6d_results/detection_ism.json')
+
+    parser.add_argument("--output_dir", nargs="?", default=default_output_dir, help="Path to root directory of the output")
+    parser.add_argument("--cad_path", nargs="?", default=default_cad_path, help="Path to CAD(mm)")
+    parser.add_argument("--rgb_path", nargs="?", default=default_rgb_path, help="Path to RGB image")
+    parser.add_argument("--depth_path", nargs="?", default=default_depth_path, help="Path to Depth image(mm)")
+    parser.add_argument("--cam_path", nargs="?", default=default_camera_path, help="Path to camera information")
+    parser.add_argument("--seg_path", nargs="?", default=default_seg_path, help="Path to segmentation information(generated by ISM)")
+    parser.add_argument("--det_score_thresh", default=0.2, help="The score threshold of detection")
```
allnes: No `type=float`, so the value stays a string; the later comparison `det['score'] > det_score_thresh` (`float > str`) raises `TypeError` when the flag is passed. Same at `pem_model_convert_ov_ir.py:127`. `run_inference_custom_openvino.py:200` already has `type=float`, so the three copies have drifted. Please add `type=float` here.

modules/3d/OV-SAM-6D/SAM-6D/Instance_Segmentation_Model/export_ism.py
```diff
@@ -0,0 +1,1189 @@
...
+    print(f'[export:dinov2_ext] {tag} IR: {xml_path} ({size_mb:.1f} MB)')
+
+    # --- Step 5: Verification ---
+    if verify:
+        core = ov.Core()
+        compiled = core.compile_model(xml_path, 'CPU')
+        N = 2
+        dummy_images = np.random.randn(N, 3, 224, 224).astype(np.float32)
+        dummy_masks = np.random.rand(N, 1, 224, 224).astype(np.float32)
+        result = compiled({'images': dummy_images, 'proposal_masks': dummy_masks})
```
allnes: The base DINOv2 IR input is never renamed to `images` (it keeps its default parameter name/index), so `--verify` raises a port-not-found / `KeyError` here. Index inputs positionally (`compiled([dummy_images, dummy_masks])`) or set input names via `set_names` when building the model.

modules/3d/OV-SAM-6D/SAM-6D/Instance_Segmentation_Model/export_ism.py
```diff
@@ -0,0 +1,1189 @@
...
+
+    class _XFormersBlocker(importlib.abc.Loader):
+        """Fake loader that always raises ImportError."""
+        def create_module(self, spec):
+            return None
+        def exec_module(self, module):
+            raise ImportError("xformers blocked by export_ism.py")
+
+    class _XFormersFinder(importlib.abc.MetaPathFinder):
+        def find_module(self, fullname, path=None):
```
allnes: `find_module` is the `MetaPathFinder` API removed in Python 3.12; on newer Python the finder is never consulted, so xformers is not actually blocked (it works only because the env pins Python 3.11.6). Please implement `find_spec(self, fullname, path=None, target=None)` instead.

modules/3d/OV-SAM-6D/SAM-6D/Instance_Segmentation_Model/infer_ism_ov.py
```diff
@@ -0,0 +1,2366 @@
...
+        rgbs = rgb.unsqueeze(0).repeat(num_proposals, 1, 1, 1)
+        masked_rgbs = rgbs * masks.unsqueeze(1)
+        return self.rgb_proposal_processor(masked_rgbs, boxes)
+
+    def process_masks_proposals(
+        self,
+        masks: torch.Tensor,
+        boxes: torch.Tensor,
+    ) -> torch.Tensor:
+        masks.unsqueeze_(1)
```
allnes: `unsqueeze_(1)` mutates the caller's `masks` tensor in place, and `.squeeze_()` on line 795 removes all singleton dims including the batch dim when `N==1`, which corrupts a single-proposal batch. Use out-of-place `masks.unsqueeze(1)` and `squeeze(1)` with an explicit dim. Same at lines 1128-1129.

modules/3d/OV-SAM-6D/SAM-6D/Instance_Segmentation_Model/infer_ism_ov.py
```diff
@@ -0,0 +1,2366 @@
...
+
+    def _compile(core, xml_path, device, model_ir=None):
+        if model_ir is None:
+            model_ir = core.read_model(xml_path)
+
+        if device.upper().startswith("GPU"):
+            if precision == "fp32":
+                _prec = "f32"
+            else:
+                _prec = "f16"
```
allnes: `init_model_ov` (base) deliberately keeps the SAM mask decoder in f32 at lines 1442-1445 (`"f32" if "sam_mask_decoder" in _name else "f16"`), but here `init_model_ov_ext` runs the decoder in f16 under the default `fp16`. Since the ext path is the recommended mode in the README, this can silently reduce accuracy. Please apply the same `sam_mask_decoder` -> f32 rule here.

modules/3d/OV-SAM-6D/SAM-6D/Instance_Segmentation_Model/infer_ism_ov.py
```diff
@@ -0,0 +1,2366 @@
...
+        processed_masks = []
+
+        for mask, box in zip(masks, boxes):
+            x0, y0, x1, y1 = box.long().tolist()
+            mask_crop = mask[y0:y1, x0:x1].float().unsqueeze(0)
+            rgb_crop = rgb[:, y0:y1, x0:x1] * mask_crop
+
+            crop_h, crop_w = mask_crop.shape[-2:]
+            scale = max(target_h, target_w) / max(crop_h, crop_w)
+            rgb_crop = F.interpolate(
```
allnes, line 1146: `F.interpolate` has no `mode=` here, so it defaults to `'nearest'` for the RGB crop, while the mask crop just below explicitly uses `mode="nearest"` and the reference pipeline resizes RGB bilinearly. Nearest-neighbor on RGB degrades the descriptor input. Please add `mode="bilinear"` (with appropriate `align_corners`) at 1148 and 1166.

modules/3d/OV-SAM-6D/SAM-6D/Pose_Estimation_Model/model/ov_pointnet2_op/custom_svd_u.cpp
```diff
@@ -0,0 +1,123 @@
...
+    size_t in_mat_size = m * n;
+    size_t u_mat_size = m * m;
+    size_t s_vec_size = std::min(m, n);
+    size_t v_mat_size = n * n;
+    
+    for (size_t b = 0; b < batch; ++b) {
+        const float* batch_data = data + b * in_mat_size;
+        Eigen::Map<const Eigen::Matrix<float, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>> A(batch_data, m, n);
+        
+        // Use Eigen's BDCSVD for better numerical stability (similar to LAPACK)
```
allnes: Two things: the comment says `BDCSVD` but the code uses `JacobiSVD`; and `ensure_svd_u_signs` only flips negative singular values and re-sorts, but `JacobiSVD` already returns non-negative, descending singular values, so the helper is effectively a no-op and does not enforce the PyTorch/LAPACK U-column sign convention. Since `U` feeds a rotation (pose) downstream, sign ambiguity vs `torch.svd` can flip results. Please align the sign convention with the PyTorch reference and fix the comment.

modules/3d/OV-SAM-6D/SAM-6D/Pose_Estimation_Model/model/ov_pointnet2_op/ball_query.cpp
```diff
@@ -0,0 +1,105 @@
+// Copyright (C) 2018-2026 Intel Corporation
```
allnes, line 1: This file, along with `furthest_point_sampling.cpp`, `grouping_operation.cpp` and `gather_operation.cpp`, is a port of the MIT-licensed PointNet2 CUDA kernels (Erik Wijmans / Charles R. Qi) but carries only the Intel Apache-2.0 header. When adapting MIT-licensed code the original copyright and permission notice must be retained. Please add an upstream attribution line to these headers, and a `third-party-programs.txt` entry if appropriate.

modules/3d/OV-SAM-6D/SAM-6D/ov_environment_u24.yaml
```diff
@@ -0,0 +1,44 @@
...
+   - hydra-colorlog
+   - hydra-core
+   - gdown
+   - pandas
+   - imageio
+   - pyrender
+   - pytorch-lightning
+   - pycocotools
+   - distinctipy
+   - git+https://github.com/facebookresearch/segment-anything.git  # SAM
```
allnes, line 35: `third-party-programs.txt` covers `ultralytics` (AGPL-3.0) and `segment-anything`, but there are no dedicated entries for DINOv2 (Apache-2.0) or FastSAM. FastSAM builds on ultralytics/YOLO (AGPL). Please confirm the AGPL exposure is acceptable for a redistributed OpenVINO module and that DINOv2/FastSAM notices are covered.

modules/3d/OV-SAM-6D/SAM-6D/Pose_Estimation_Model/eval_utils.py
```diff
@@ -0,0 +1,470 @@
...
+
+    depth_gt = _render_depth(renderer, mesh_pyrender, K, R_gt, t_gt, w, h)
+    dist_gt = _depth_to_distance(depth_gt, K)
+
+    depth_est = _render_depth(renderer, mesh_pyrender, K, R_est, t_est, w, h)
+    dist_est = _depth_to_distance(depth_est, K)
+
+    visib_gt = (dist_gt > 0) & ((dist_test == 0) | (dist_test >= dist_gt - delta))
+    visib_est = (dist_est > 0) & ((dist_test == 0) | (dist_test >= dist_est - delta))
+    visib_est = visib_est | (visib_gt & (dist_est > 0))
```
allnes: This mixes GT visibility into the estimated-visibility mask. Standard BOP VSD defines `visib_est` purely from `dist_est`/`dist_test`; augmenting it with `visib_gt` biases `visib_union`/`visib_inter` and therefore `AR_VSD`. Please verify against the BOP reference; if unintended, remove this line.

modules/3d/OV-SAM-6D/SAM-6D/Pose_Estimation_Model/run_inference_custom_openvino.py
```diff
@@ -0,0 +1,645 @@
...
+    radius = np.max(np.linalg.norm(model_points, axis=1))
+
+    all_rgb, all_cloud, all_rgb_choose, all_score, all_dets = [], [], [], [], []
+    for inst in dets:
+        seg = inst['segmentation']
+        score = inst['score']
+        h, w = seg['size']
+        try:
+            rle = cocomask.frPyObjects(seg, h, w)
+        except:
```
allnes: Bare `except:` swallows `KeyboardInterrupt`/`SystemExit` and hides real errors. Please use `except Exception:` (or narrow to the pycocotools error). Same at `run_inference_custom_pytorch.py:315` and `pem_model_convert_ov_ir.py:309`.

modules/3d/OV-SAM-6D/SAM-6D/Pose_Estimation_Model/pem_model_convert_ov_ir.py
```diff
@@ -0,0 +1,607 @@
...
+            export_params=True,
+            keep_initializers_as_inputs=False
+        )
+        print(f"[ONNX] PEM onnx model export success: {onnx_model_path}")
+            
+    except Exception as e:
+        print(f"[ONNX] export failed : {e}")
+
+
+def openvino_model_convert_feature_extraction_submodel(core, ov_fe_input_name, ov_fe_output_name, ov_fe_input, onnx_fe_model_path, ov_fe_model_path):
```
allnes: `ov_fe_input_name`/`ov_fe_output_name`/`ov_fe_input` are accepted but never used, and `compiled_model` (line 517) is computed then discarded, so the FE model is saved as-is with no reshape applied. Either apply the intended input reshaping or drop the dead compile and the unused parameters. Same dead `compiled_model` at line 527.

modules/3d/OV-SAM-6D/SAM-6D/Pose_Estimation_Model/pem_model_convert_ov_ir.py
```diff
@@ -0,0 +1,607 @@
...
+    ov_pem_input_name, ov_pem_output_name, ov_pem_input, \
+    ov_fe_input_name, ov_fe_output_name, ov_fe_input = prepare_data(cfg, model, device)
+
+    model_save_path = "model_save"
+    os.makedirs(model_save_path, exist_ok=True)
+    onnx_pem_model_path = os.path.join(model_save_path, 'onnx_pem_model_cpu.onnx')
+    onnx_fe_model_path = os.path.join(model_save_path, 'onnx_fe_model_cpu.onnx')
+    ov_pem_model_path = os.path.join(model_save_path, 'ov_pem_model_cpu.xml')
+    ov_fe_model_path = os.path.join(model_save_path, 'ov_fe_model_cpu.xml')
+    ov_extension_lib_path = './model/ov_pointnet2_op/build/libopenvino_operation_extension.so'
```
allnes: Relative hardcoded path with no `os.path.isfile` check before `core.add_extension` (line 595), which fails opaquely from a different CWD or when the extension is not built. Please resolve relative to `__file__`/`BASE_DIR` and check existence with a clear error. Same at `run_inference_custom_openvino.py:414`.

modules/3d/OV-SAM-6D/SAM-6D/Pose_Estimation_Model/run_inference_custom_pytorch.py
```diff
@@ -0,0 +1,484 @@
...
+    # concat side by side in PIL
+    rgb = Image.fromarray(np.uint8(rgb))
+    img = np.array(img)
+    concat = Image.new('RGB', (img.shape[1] + prediction.size[0], img.shape[0]))
+    concat.paste(rgb, (0, 0))
+    concat.paste(prediction, (img.shape[1], 0))
+    return concat
+
+
+def _get_template(path, cfg, device, tem_index=1):
```
allnes, line 143: `_get_template`/`get_templates`/`get_test_data` here are byte-for-byte identical to `pem_model_convert_ov_ir.py:179-361`, and `run_inference_custom_openvino.py:258-397` is a NumPy port of the same. The `Config` class + `init()` and `visualize()` are also duplicated across all three files. This is the source of the drift bugs above (`det_score_thresh` typing, pose-score handling). Consider extracting a shared module parameterized by backend.

modules/3d/OV-SAM-6D/SAM-6D/Instance_Segmentation_Model/infer_ism_ov.py
```diff
@@ -0,0 +1,2366 @@
...
+
+    proposal_processor = CropResizePad(224)
+
+    print(
+        f"OV ISM pipeline ready  (device={ov_device})"
+    )
+    return model, cfg, torch_device, selected_poses, proposal_processor
+
+
+def init_model_ov_ext(
```
allnes, line 1598: `init_model_ov_ext` is about 85% identical to `init_model_ov` (1380-1561): hydra config load, the `_compile` closure, fastsam branch, DINOv2/mask-generator construction, pose-table tail. The only meaningful divergence is decoder precision (see the f16 comment above), which is the kind of drift duplication causes. Consider extracting shared helpers.

modules/3d/OV-SAM-6D/SAM-6D/Instance_Segmentation_Model/export_ism.py
```diff
@@ -0,0 +1,1189 @@
...
+        _verify_model(wrapper, ov_model, dummy, "DINOv2 ViT-L/14")
+
+    return xml_path
+
+
+# ===================================================================
+#  Extended export: bake pre/post-processing custom ops into IR
+# ===================================================================
+
+def export_sam_encoder_ext(sam, output_dir, compress_fp16=False,
```
allnes, line 446: `export_sam_encoder_ext`/`export_sam_decoder_ext`/`export_sam_decoder_ext_v2` share large copy-pasted blocks: reuse-base-IR, the interpolate/crop/interpolate postprocess (539-568 and 669-697 are byte-identical), and the save/size-print tail. `export_sam_decoder_ext` is nearly a subset of `_v2`. Consider consolidating the shared graph-build and save logic.

modules/3d/OV-SAM-6D/SAM-6D/Pose_Estimation_Model/test_bop_subset_eval_ov.py
```diff
@@ -0,0 +1,610 @@
...
+    return cfg
+
+
+def to_numpy(x):
+    if isinstance(x, torch.Tensor):
+        return x.detach().cpu().numpy()
+    return np.asarray(x)
+
+
+def project_points(K: np.ndarray, pts_mm: np.ndarray) -> np.ndarray:
```
allnes: `compute_pose_errors`/`transform_points`/`project_points` here are self-described "approximate, without symmetry handling" duplicates of the symmetry-aware `compute_mssd`/`compute_mspd` in `eval_utils.py`. Two divergent metric implementations produce different AR numbers between scripts. Consider reusing `eval_utils.py`.

modules/3d/OV-SAM-6D/SAM-6D/Pose_Estimation_Model/test_bop_subset_eval_ov.py
```diff
@@ -0,0 +1,610 @@
...
+    t0 = time.time()
+    ov_fe_compiled, ov_pem_compiled = init_openvino_models(args)
+    print(f"OpenVINO init time: {time.time() - t0:.2f}s")
+
+    # Monkey-patch load_objs to filter out non-directory entries (e.g. .tar.gz
+    # files that may sit alongside obj_XXXXXX template folders).
+    import utils.bop_object_utils as _bop_obj_utils
+    _orig_load_objs = _bop_obj_utils.load_objs
+
+    def _load_objs_filtered(model_path='models', template_path='render_imgs',
```
allnes: Replacing the module-global `glob.glob` is not thread-safe and affects any concurrent glob in the process; the `finally` restores it but the approach is fragile. Consider filtering results after calling `load_objs`, or passing a directory list instead of patching a stdlib function.

modules/3d/OV-SAM-6D/OV_README.md
```diff
@@ -0,0 +1,347 @@
...
+- Intel GPU drivers installed (for GPU inference)
+- Python 3.10+
+- Internet connection (required for model and Blender downloads)
+
+---
+
+# 1. Create and Activate Conda Environment
+
+```bash
+cd JiehongLin-SAM-6D
```
allnes: Leftover from the original repo naming: every other path here uses `openvino_contrib/modules/3d/OV-SAM-6D/...`. Should be `cd openvino_contrib/modules/3d/OV-SAM-6D/SAM-6D`.

modules/3d/OV-SAM-6D/OV_README.md
```diff
@@ -0,0 +1,347 @@
+<!--
+Copyright (C) 2018-2026 Intel Corporation
+SPDX-License-Identifier: Apache-2.0
+-->
+
+# Note
+
+_This directory is an OpenVINO port of [SAM-6D](https://github.com/JiehongLin/SAM-6D) (based on commit 1c2543b). The original SAM-6D source files are **not** committed here — they are fetched and patched by `setup_from_original.sh`._
```
allnes, line 8: The original-repo and commit-id attribution should also live in a top-level `modules/3d/OV-SAM-6D/README.md` (the module root the top-level repo README links to), not only in this nested `OV_README.md`. Please add a short module-root README with this attribution.

modules/3d/OV-SAM-6D/OV_README.md
```diff
@@ -0,0 +1,347 @@
...
+# 2. Install OpenVINO Nightly Build
+
+Download the OpenVINO nightly build package from:
+
+https://storage.openvinotoolkit.org/repositories/openvino/packages/nightly/
+
+Download:
+
+```text
+openvino_toolkit_ubuntu24_2026.3.0.dev20260519_x86_64.tgz
```
allnes, line 102: A specific nightly build (`dev20260519`) is hardcoded here and at lines 73/79/319. Nightly artifacts are pruned over time, so these instructions will break. Reference a stable release or a version range, and note that it should be replaced with the current nightly.

modules/3d/OV-SAM-6D/SAM-6D/Pose_Estimation_Model/model/ov_pointnet2_op/CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
+# Copyright (C) 2018-2026 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+# [cmake:extension]
+set(CMAKE_CXX_STANDARD 11)
```
allnes: The extension uses `ov::Op`/`AttributeVisitor` and links `openvino::runtime`, which is built with C++17. Compiling the extension at C++11 risks ABI/warning mismatches. Recommend `set(CMAKE_CXX_STANDARD 17)`.

modules/3d/OV-SAM-6D/SAM-6D/Pose_Estimation_Model/pem_model_convert_ov_ir.py
```diff
@@ -0,0 +1,607 @@
...
+from openvino import Core
+
+import pycocotools.mask as cocomask
+import trimesh
+
+from utils.data_utils import load_im, get_bbox, get_point_cloud_from_depth, get_resize_rgb_choose
+from utils.draw_utils import draw_detections
+import xml.etree.ElementTree as ET
+
+import onnx
```
allnes: `numpy` is re-imported (already at line 9); `onnx`, `helper` and `numpy_helper` (29-31) are never used; `from PIL import Image` (line 7) is also unused. Please remove them.

modules/3d/OV-SAM-6D/SAM-6D/Instance_Segmentation_Model/export_ism.py
```diff
@@ -0,0 +1,1189 @@
...
+
+    from model.dinov2 import _make_dinov2_model, descriptor_map
+
+    arch = descriptor_map[model_name]
+    ckpt = os.path.join(checkpoint_dir, f"{model_name}_pretrain.pth")
+
+    print(f"[export] Loading DINOv2 ({model_name}, arch={arch}) from {ckpt}")
+
+    model = _make_dinov2_model(arch_name=arch, pretrained=False)
+    state = torch.load(ckpt, map_location="cpu")
```
allnes: No `os.path.isfile(ckpt)` check before loading (opaque failure on a missing checkpoint), and `weights_only=` is unset (recent torch warns and will change the default). Please add an existence check and pass `weights_only=True`.

modules/3d/OV-SAM-6D/SAM-6D/Instance_Segmentation_Model/export_ism.py
```diff
@@ -0,0 +1,1189 @@
...
+
+    if (args.ext and export_all) or args.only == "sam_decoder_ext_v2":
+        if sam is None:
+            sam = build_sam(args.sam_checkpoint_dir, args.sam_model_type)
+        results["sam_decoder_ext_v2"] = export_sam_decoder_ext_v2(
+            sam, args.output_dir, compress_fp16=args.fp16,
+            verify=args.verify)
+
+    del sam
+    torch.cuda.empty_cache() if torch.cuda.is_available() else None
```
allnes: A ternary expression used only for its side effect is hard to read. Prefer: `if torch.cuda.is_available(): torch.cuda.empty_cache()`

modules/3d/OV-SAM-6D/SAM-6D/Pose_Estimation_Model/run_inference_custom_pytorch.py
```diff
@@ -0,0 +1,484 @@
...
+    # pytorch save results
+    if 'pred_pose_score' in input_data.keys():
+        pose_scores = pred_pose_score * input_data['score']
+    else:
+        pose_scores = input_data['score']
+    pose_scores = pose_scores.detach().cpu().numpy()
+    pred_rot = pred_R
+    pred_trans = pred_t * 1000
+
+    print("[PyTorch] visualizating ...")
```
allnes: This log line is printed twice back-to-back (also at line 477); the first precedes results-saving, not visualization, and there is a typo ("visualizating"). Please remove the duplicate and fix the wording.

allnes (changes requested):

modules/3d/OV-SAM-6D/SAM-6D/Instance_Segmentation_Model/run_ism.sh
pravin25: done

modules/3d/OV-SAM-6D/SAM-6D/Pose_Estimation_Model/run_inference_custom_pytorch.py
pravin25: done

modules/3d/OV-SAM-6D/SAM-6D/Pose_Estimation_Model/run_inference_custom_pytorch.py
pravin25: done

modules/3d/OV-SAM-6D/SAM-6D/Instance_Segmentation_Model/export_ism.py
pravin25: done

modules/3d/OV-SAM-6D/SAM-6D/Instance_Segmentation_Model/export_ism.py
pravin25: done

modules/3d/OV-SAM-6D/SAM-6D/Instance_Segmentation_Model/infer_ism_ov.py
pravin25: done

modules/3d/OV-SAM-6D/SAM-6D/Instance_Segmentation_Model/infer_ism_ov.py
pravin25: Done. I aligned init_model_ov_ext() with init_model_ov() for GPU compilation: in the default fp16 mode, sam_mask_decoder IRs are still compiled with f32 precision hints, while the remaining models use f16. When precision="fp32", all models are compiled with f32.

modules/3d/OV-SAM-6D/SAM-6D/Instance_Segmentation_Model/infer_ism_ov.py
pravin25: done

modules/3d/OV-SAM-6D/SAM-6D/Pose_Estimation_Model/model/ov_pointnet2_op/custom_svd_u.cpp
pravin25: done

modules/3d/OV-SAM-6D/SAM-6D/Pose_Estimation_Model/model/ov_pointnet2_op/ball_query.cpp
pravin25: done

modules/3d/OV-SAM-6D/SAM-6D/ov_environment_u24.yaml
pravin25: done. I added dedicated DINOv2 and FastSAM entries to third-party-programs.txt, and I made the FastSAM/ultralytics AGPL-3.0 dependency path explicit in ov_environment_u24.yaml. The remaining question of whether AGPL exposure is acceptable for redistribution is a release/legal decision rather than a code change.

modules/3d/OV-SAM-6D/SAM-6D/Pose_Estimation_Model/eval_utils.py
pravin25: done, I verified and removed the GT-to-estimated visibility merge in compute_vsd(). visib_est is now defined purely from dist_est/dist_test, and visib_union/visib_inter are computed from standard BOP masks only.

modules/3d/OV-SAM-6D/SAM-6D/Pose_Estimation_Model/run_inference_custom_openvino.py
pravin25: done

modules/3d/OV-SAM-6D/SAM-6D/Pose_Estimation_Model/pem_model_convert_ov_ir.py
pravin25: done

modules/3d/OV-SAM-6D/SAM-6D/Pose_Estimation_Model/run_inference_custom_pytorch.py
pravin25: done,

modules/3d/OV-SAM-6D/SAM-6D/Pose_Estimation_Model/pem_model_convert_ov_ir.py
pravin25: done, ov_pem_output_name was unused in the OpenVINO PEM conversion path, so I removed it from the function signature and call, and cleaned up the corresponding dead variable in prepare_data() return/unpack.

modules/3d/OV-SAM-6D/SAM-6D/Instance_Segmentation_Model/infer_ism_ov.py
pravin25: done

modules/3d/OV-SAM-6D/SAM-6D/Instance_Segmentation_Model/export_ism.py
pravin25: done,  I refactored the SAM ext export path to reduce duplication:
Added shared helpers for base-IR reuse and save/size-print logic.
Extracted the common decoder postprocess subgraph into one builder used by both export_sam_decoder_ext and export_sam_decoder_ext_v2.
Kept only v2-specific parts in export_sam_decoder_ext_v2 (stability score and mask-to-box graph).
This should reduce drift risk while preserving behavior.

modules/3d/OV-SAM-6D/SAM-6D/Pose_Estimation_Model/test_bop_subset_eval_ov.py
pravin25: done, this comment has been addressed by reusing symmetry-aware compute_mssd/compute_mspd from eval_utils.py in the subset evaluator and removing the duplicate approximate local implementations.

modules/3d/OV-SAM-6D/SAM-6D/Pose_Estimation_Model/test_bop_subset_eval_ov.py
pravin25: done

modules/3d/OV-SAM-6D/OV_README.md
pravin25: done

modules/3d/OV-SAM-6D/OV_README.md
pravin25: done

modules/3d/OV-SAM-6D/OV_README.md
pravin25: currently we are using nightly build because, in ov gpu plugin I submitted patches for custom ops memory support, those patches not yet part of OpenVino 2026.2.1  release, it will come in OpenVino 2026.3 release, thats why i am using currently nightly build, in future we can change this instruction

modules/3d/OV-SAM-6D/SAM-6D/Pose_Estimation_Model/model/ov_pointnet2_op/CMakeLists.txt
pravin25: done

modules/3d/OV-SAM-6D/SAM-6D/Pose_Estimation_Model/pem_model_convert_ov_ir.py
pravin25: done

modules/3d/OV-SAM-6D/SAM-6D/Instance_Segmentation_Model/export_ism.py
pravin25: done

modules/3d/OV-SAM-6D/SAM-6D/Instance_Segmentation_Model/export_ism.py
pravin25: done

modules/3d/OV-SAM-6D/SAM-6D/Pose_Estimation_Model/run_inference_custom_pytorch.py
pravin25: done

modules/3d/OV-SAM-6D/SAM-6D/Pose_Estimation_Model/run_inference_custom_openvino.py
```diff
@@ -0,0 +1,506 @@
...
+    cfg = init()
+    _seed_everything(cfg.seed)
+    precision_mode = cfg.precision
+    print(f"[OpenVINO] Device: {cfg.device}, Precision: {precision_mode}, Seed: {cfg.seed}")
+
+    # Paths
+    ov_pem_model_path = "model_save/ov_pem_model_cpu.xml"
+    ov_fe_model_path = "model_save/ov_fe_model_cpu.xml"
+    ov_gpu_kernel_path = "./model/ov_pointnet2_op/pem_gpu_ops.xml"
+    ov_extension_lib_path = "./model/ov_pointnet2_op/build/libopenvino_operation_extension.so"
```
allnes: This was marked done but is not fixed here. The extension path is still relative and passed to `core.add_extension` (line 279) with no existence check and no resolution against `BASE_DIR` (already defined at line 30). Launched from a different CWD, or before the `.so` is built, this fails with an opaque OpenVINO error.

`test_bop_subset_eval_ov.py:264` already implements the requested pattern — please apply the same here:

```python
ov_extension_lib_path = os.path.join(BASE_DIR, "model/ov_pointnet2_op/build/libopenvino_operation_extension.so")
if not os.path.isfile(ov_extension_lib_path):
    raise FileNotFoundError(f"Extension not found: {ov_extension_lib_path}")
```

Same at `pem_model_convert_ov_ir.py:399`.

modules/3d/OV-SAM-6D/SAM-6D/Pose_Estimation_Model/common_infer_utils.py
```diff
@@ -0,0 +1,287 @@
...
+        xyz = np.load(xyz_path).astype(np.float32) / 1000.0
+        mask = cv2.imread(mask_path, cv2.IMREAD_GRAYSCALE).astype(np.uint8) == 255
+
+        rows = np.any(mask, axis=1)
+        cols = np.any(mask, axis=0)
+        y1, y2 = np.where(rows)[0][[0, -1]]
+        x1, x2 = np.where(cols)[0][[0, -1]]
+        mask = mask[y1:y2, x1:x2]
+
+        rgb = rgb[:, :, ::-1][y1:y2, x1:x2, :]
```
allnes, line 84: Channel-order mismatch between template and query in the OpenVINO path.

The numpy branch loads with `cv2.imread` (BGR, line 74) then flips `[:, :, ::-1]` → **RGB**. But the torch branch and the observed image use `load_im` (RGB, line 107) then flip → **BGR**. So in the OV pipeline the templates are fed as RGB while the query crop is BGR, and the numpy template is R/B-swapped relative to the torch template.

This systematically degrades ViT feature matching and pose accuracy on the recommended OV mode. Please use `load_im` in the numpy branch too (or drop the flip) so both paths match.

modules/3d/OV-SAM-6D/SAM-6D/Pose_Estimation_Model/model/ov_pointnet2_op/pem_gpu_ops.xml
```diff
@@ -0,0 +1,102 @@
+<!-- Copyright (C) 2018-2026 Intel Corporation
+SPDX-License-Identifier: Apache-2.0 -->
+
+<CustomLayer name="FurthestPointSampling" type="SimpleGPU" version="1">
+    <Kernel entry="ov_furthest_point_sampling">
+	    <Source filename="furthest_point_sampling.cl"/>
+        <!-- <Define name="npoint" type="int"   param="npoint"/> -->
+        <!-- <Define name="MAX_NPOINT" type="int" param="max_npoint" default="196"/> -->
+    </Kernel>
+    <Buffers>
```
allnes, line 10: The PR's own `validate_gpu_ops_xml.py` fails on this shipped XML. Running it reports 8 errors: the input tensors of `FurthestPointSampling`, `GatherOperation`, `BallQuery` and `GroupingOperation` have `format=''` where the validator expects `format="any"`.

The validator was written specifically to prevent reorder naming collisions. Either the XML is wrong (real GPU risk) or the validator is too strict — please reconcile them.

modules/3d/OV-SAM-6D/SAM-6D/Pose_Estimation_Model/model/ov_pointnet2_op/furthest_point_sampling.cl
```diff
@@ -0,0 +1,152 @@
...
+    // First sampled point = 0; cache in local memory
+    if (lid == 0) {
+        output[out_offset] = 0;
+        s_idx[0] = 0;
+    }
+    barrier(CLK_LOCAL_MEM_FENCE);
+
+    for (int j = 1; j < npoint; ++j) {
+        // Read last selected index from LOCAL memory
+        int last_idx = s_idx[0];
```
allnes, line 50: Possible race on `s_idx[0]`. All work-items read `last_idx = s_idx[0]` at the start of each iteration, but `s_idx[lid]` (including `lid == 0`, i.e. `s_idx[0]`) is overwritten at line 135 with no barrier between the read and that write.

With a 1024-wide workgroup spanning multiple subgroups, `lid 0` can overwrite `s_idx[0]` before another work-item reads it, corrupting the selected index. This looks like the barrier removed in the optimization commit. Please add a `barrier(CLK_LOCAL_MEM_FENCE)` after reading `last_idx`, or keep it in a separate local variable not reused by the reduction. (Not verified on hardware — may not surface on PTL.)

modules/3d/OV-SAM-6D/SAM-6D/Instance_Segmentation_Model/infer_ism_ov.py
```diff
@@ -0,0 +1,2341 @@
...
+
+        for p in (sam_encoder_xml, sam_decoder_xml):
+            if not os.path.isfile(p):
+                raise FileNotFoundError(
+                    f"Extended OV IR not found: {p}  "
+                    f"(run export_ism.py --ext first)")
+
+        encoder_compiled = _compile(sam_encoder_xml)
+
+        _ppb = int(cfg.model.segmentor_model.points_per_batch)
```
allnes, line 1649: The decoder IR is statically reshaped so batch == `points_per_batch`, but the final point chunk is smaller when `points_per_side**2 % points_per_batch != 0`, causing a shape mismatch at inference. Safe for the defaults (32² = 1024, ppb = 64), but crashes for e.g. `--points_per_side 30`. Please keep the batch dim dynamic, or pad the last chunk.

modules/3d/OV-SAM-6D/SAM-6D/Pose_Estimation_Model/pem_model_convert_ov_ir.py
```diff
@@ -0,0 +1,414 @@
+# Copyright (C) 2018-2026 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+import argparse
+import os
+import sys
+import os.path as osp
+import numpy as np
```
allnes: Dead code remaining after the unused-imports / dead-vars fixes were marked done: `import numpy as np` is never used in this file. Also `ov_fe_input_name` / `ov_fe_output_name` / `ov_fe_input` are still built in `prepare_data` (lines 240-254) and unpacked at line 391, but never used afterwards (the FE-convert function no longer takes them). Please remove.

modules/3d/OV-SAM-6D/SAM-6D/Instance_Segmentation_Model/export_ism.py
```diff
@@ -0,0 +1,1191 @@
...
+# ===================================================================
+def _block_xformers():
+    """
+        Block xformers import BEFORE any model code is loaded.
+    """
+    if "xformers" in sys.modules:
+        _patch_xformers_flags()
+        return
+
+    import importlib
```
allnes, line 53: `_block_xformers` does `import importlib` but then uses `importlib.abc.MetaPathFinder` (line 62) and `importlib.util.spec_from_loader`. On Python 3.12 a bare `import importlib` does not bind the `abc` / `util` submodules; this only works today as a side effect of `import torch` at the top of the module.

Since this is the documented recommended path (`ISM_BLOCK_XFORMERS=1`), please `import importlib.abc, importlib.util` inside the function.

modules/3d/OV-SAM-6D/SAM-6D/Pose_Estimation_Model/test_bop_subset_eval_ov.py
```diff
@@ -0,0 +1,602 @@
...
+                    mspd_px = float(best[2])
+                    vsd_mm = float(best[3])
+
+                    mssd_abs_thresh = [r * diameter_mm for r in mssd_thresh_ratio]
+                    vsd_abs_thresh = [r * diameter_mm for r in vsd_thresh_ratio]
+
+                    ar_mssd = ar_from_thresholds(mssd_mm, mssd_abs_thresh)
+                    ar_mspd = ar_from_thresholds(mspd_px, mspd_thresh_px)
+                    ar_vsd = ar_from_thresholds(vsd_mm, vsd_abs_thresh)
+                    ar_mean = float(np.mean([ar_mssd, ar_mspd, ar_vsd]))
```
allnes: `ar_vsd` here comes from `compute_vsd_surrogate` (line 137) = `mean(|Δz|)`, with no rendering, no visibility mask and no symmetry handling. Folding this surrogate into `ar_mean` (and the summary) makes the reported AR not comparable to BOP AR. At minimum, please do not average the surrogate into `AR_mean`.

modules/3d/OV-SAM-6D/SAM-6D/Pose_Estimation_Model/run_inference_custom_pytorch.py
```diff
@@ -0,0 +1,274 @@
...
+        det_score_thresh,
+        cfg,
+        device,
+    )
+
+
+if __name__ == "__main__":
+    cfg = init()
+
+    random.seed(cfg.rd_seed)
```
allnes, line 167: Only `random` and `torch` are seeded here, but template/observed point sampling uses `np.random.choice` in `common_infer_utils`. Without `np.random.seed(...)` the runs are non-deterministic and the PyTorch-vs-OpenVINO comparison is not apples-to-apples. `run_inference_custom_openvino.py` and `test_bop_subset_eval_ov.py` correctly seed numpy — please add `np.random.seed(cfg.rd_seed)` here and in `pem_model_convert_ov_ir.py:341`.

modules/3d/OV-SAM-6D/revert_setup.sh
```diff
@@ -0,0 +1,112 @@
...
+Instance_Segmentation_Model/infer_ism_ov.py
+Instance_Segmentation_Model/ov_sam_layer_norm.py
+Instance_Segmentation_Model/run_ism.sh
+ov_environment_u24.yaml
+patches/ism.patch
+patches/pem.patch
+Pose_Estimation_Model/eval_utils.py
+Pose_Estimation_Model/common_infer_utils.py
+Pose_Estimation_Model/run_pos.sh
+Pose_Estimation_Model/.gitattributes
```
allnes: Typo in `KEEP_LIST`: `Pose_Estimation_Model/.gitattributes` does not exist — the committed file is `Pose_Estimation_Model/.gitignore`. As a result `revert_setup.sh` will delete the tracked `.gitignore`. Please fix the entry to `.gitignore`.

modules/3d/OV-SAM-6D/OV_README.md
```diff
@@ -0,0 +1,343 @@
...
+- Intel GPU drivers installed (for GPU inference)
+- Python 3.10+
+- Internet connection (required for model and Blender downloads)
+
+---
+
+# 1. Create and Activate Conda Environment
+
+```bash
+cd openvino_contrib/modules/3d/OV-SAM-6D/SAM-6D
```
allnes: Broken double-`cd` (regression from the previous path fix): line 49 already `cd`s into `.../OV-SAM-6D/SAM-6D`, then line 50 does `cd SAM-6D`, which does not exist. Please drop line 50. There is also trailing whitespace on line 49.

modules/3d/OV-SAM-6D/SAM-6D/Pose_Estimation_Model/model/ov_pointnet2_op/custom_svd_u.cl
```diff
@@ -0,0 +1,266 @@
...
+    float maxDiagEntry = fmax(fabs(W[0][0]), fmax(fabs(W[1][1]), fabs(W[2][2])));
+
+    // Compute threshold like Eigen
+    //float considerAsZero = 1.17549435e-38f;        // FLT_MIN
+    float considerAsZero = 1.17549435e-7f;        // FLT_MIN
+    float precision      = 2.0f * 1.1920929e-07f; // 2*FLT_EPSILON
+    float threshold      = fmax(considerAsZero, precision * maxDiagEntry);
+
+    int pairs[3][2] = { {1,0}, {2,0}, {2,1} };
+    int max_sweeps = 10;
```
allnes, line 165: `max_sweeps` (and `cnt`, line 168) are declared but never used — the Jacobi `while (!finished)` loop (line 169) has no hard iteration cap. For well-conditioned 3×3 matrices convergence is fine, but a hard cap would guard against a GPU hang on pathological input. Same in `custom_svd_v.cl`.

modules/3d/OV-SAM-6D/SAM-6D/Instance_Segmentation_Model/eval_utils.py
```diff
@@ -0,0 +1,409 @@
...
+def _mask_iou(mask_a: np.ndarray, mask_b: np.ndarray) -> float:
+    """Compute IoU between two binary masks."""
+
+    intersection = np.logical_and(mask_a, mask_b).sum()
+    union = np.logical_or(mask_a, mask_b).sum()
+
+    return float(intersection / union) if union > 0 else 0.0
+
+
+def _compute_ap_101(recall: np.ndarray, precision: np.ndarray) -> float:
```
allnes: Dead code: `_compute_ap_101` is never referenced (only `_compute_ap_voc` is used, including inside `compute_map_iou`). The `compute_map_iou` name suggests COCO-101, but it actually uses the VOC envelope AP.

modules/3d/OV-SAM-6D/SAM-6D/Instance_Segmentation_Model/run_ism.sh
```diff
@@ -0,0 +1,207 @@
...
+  log "Segmentor model: $SEGMENTOR_MODEL"
+  log "Max images: $MAX_IMAGES"
+  log "OV_DEVICE=$OV_DEVICE OV_SAM_DEVICE=$OV_SAM_DEVICE"
+  log "Log file: $run_log"
+
+  # Run inference in background, write output to log file directly
+  # so the log is always written even if stdout is piped through head/grep.
+  local rc=0
+  OV_DEVICE="$OV_DEVICE" \
+  OV_SAM_DEVICE="$OV_SAM_DEVICE" \
```
allnes: `OV_SAM_DEVICE` is exported here but never read by `eval_ism_ov_bop.py` / `infer_ism_ov.py` (only `OV_DEVICE` → `--ov_device` is used). Dead environment variable.

third-party-programs.txt
```diff
@@ -5587,3 +5611,3078 @@ easydict
...
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+
+--------------------------------------------------------------
+
+Eigen
```
allnes, line 9972: The `Eigen` entry contains the **Apache License 2.0** text, but Eigen is licensed under **MPL-2.0** (with minor historical LGPL/BSD parts). Eigen is compiled into the extension `.so` (`CMakeLists.txt:22,29`, header-only). The `.so` is not committed, so redistribution risk is low, but the license entry must state MPL-2.0. Please replace the license text.

third-party-programs.txt
```diff
@@ -221,6 +221,30 @@ opencv
...
 
+DINOv2
+
+Apache License 2.0
+
+Copyright (c) Meta Platforms, Inc. and affiliates.
+
+This module vendors and uses DINOv2 code and model assets under Apache-2.0.
+The full Apache License 2.0 text is included in the `segment-anything`
+entry above.
```
allnes: Broken cross-reference. This DINOv2 entry points to the `segment-anything` entry "above", but segment-anything is at line 7197 (below). The FastSAM entry (line 244) has the same problem: `ultralytics` "above" is at line 7403 (below). Please change "above" → "below", or inline the license text.

modules/3d/OV-SAM-6D/SAM-6D/Pose_Estimation_Model/model/ov_pointnet2_op/ball_query.cl
```diff
@@ -0,0 +1,54 @@
+// Copyright (C) 2018-2026 Intel Corporation
```
allnes, line 1: The `.cpp` wrappers carry a "Portions adapted from the MIT-licensed PointNet2 kernels (Wijmans / Qi)" line, but the `.cl` kernels — which are the actual ports of the CUDA kernels — carry only the Intel/Apache header. Please add the same MIT attribution to at least `ball_query.cl` and `furthest_point_sampling.cl` (and the grouping/gather `.cl` for consistency).

modules/3d/OV-SAM-6D/SAM-6D/Data/Example/gt_pose.json
```diff
@@ -0,0 +1,22 @@
+[
```
allnes, line 1: These committed files (`gt_pose.json`, `models_info.json`, `mask_visib/000003_000001.png`) are derivatives of the BOP LineMOD-Occlusion dataset (GT poses / visible masks / models info) but carry no source or license note. BOP datasets have per-dataset licenses (LM-O is commonly CC BY 4.0 / research terms). Please add an attribution note (dataset name, source URL, license) near `Data/Example`, and confirm the redistribution terms permit committing these derivatives.

allnes:
### Open questions

1. **Extension path fix** — the "resolve the `.so` path against `BASE_DIR` + `os.path.isfile` check" comment was marked done, but only `test_bop_subset_eval_ov.py` was updated; `run_inference_custom_openvino.py:275` and `pem_model_convert_ov_ir.py:399` still use a relative path with no check. Was this intended, or is a follow-up planned?

2. **RGB vs BGR** — can you confirm the intended channel order? In the OpenVINO path templates end up as RGB (`cv2.imread` + flip) while the query crop ends up as BGR (`load_im` + flip), which looks inconsistent between the two branches.

3. **`validate_gpu_ops_xml.py` vs `pem_gpu_ops.xml`** — the shipped validator fails on the shipped XML (8 errors, missing `format="any"` on multi-input ops). Is the validator outdated, or does the XML need `format="any"` on those input ports for correct GPU behavior?

4. **`furthest_point_sampling.cl`** — is `s_idx[0]` guaranteed race-free between the read at line 48 and the overwrite at line 135 on a multi-subgroup 1024-wide workgroup? Was this the barrier removed in the optimization commit?

5. **Eigen license / BOP data** — will the Eigen entry in `third-party-programs.txt` be corrected to MPL-2.0, and can a source/license note be added for the committed BOP LineMOD-Occlusion example data (`Data/Example`)?

6. **Copyleft / non-commercial assets** — has there been a maintainer/legal sign-off on: (a) shipping Apache-2.0 code that imports and documents FastSAM/ultralytics (AGPL-3.0) as a first-class path, and (b) the fetched MAE backbone weights (CC-BY-NC 4.0, non-commercial) in the runtime pipeline? Confirming the commercial-use posture here would help.

modules/3d/OV-SAM-6D/SAM-6D/Pose_Estimation_Model/run_inference_custom_openvino.py
pravin25: done

modules/3d/OV-SAM-6D/SAM-6D/Pose_Estimation_Model/model/ov_pointnet2_op/pem_gpu_ops.xml
pravin25: done

modules/3d/OV-SAM-6D/SAM-6D/Pose_Estimation_Model/model/ov_pointnet2_op/furthest_point_sampling.cl
pravin25: good catch :),  done

modules/3d/OV-SAM-6D/SAM-6D/Instance_Segmentation_Model/infer_ism_ov.py
pravin25: done

modules/3d/OV-SAM-6D/SAM-6D/Pose_Estimation_Model/pem_model_convert_ov_ir.py
pravin25: done

modules/3d/OV-SAM-6D/SAM-6D/Instance_Segmentation_Model/export_ism.py
pravin25: done

modules/3d/OV-SAM-6D/SAM-6D/Pose_Estimation_Model/test_bop_subset_eval_ov.py
pravin25: done

modules/3d/OV-SAM-6D/SAM-6D/Pose_Estimation_Model/run_inference_custom_pytorch.py
pravin25: done

modules/3d/OV-SAM-6D/revert_setup.sh
pravin25: done

modules/3d/OV-SAM-6D/OV_README.md
pravin25: done

modules/3d/OV-SAM-6D/SAM-6D/Pose_Estimation_Model/model/ov_pointnet2_op/custom_svd_u.cl
pravin25: done

modules/3d/OV-SAM-6D/SAM-6D/Instance_Segmentation_Model/eval_utils.py
pravin25: done, removed dead _compute_ap_101, renamed the implementation to compute_map_iou_voc to match the actual VOC-envelope AP behavior, and kept compute_map_iou as a backward-compatible alias. Also switched internal usage to the explicit VOC-named function.

modules/3d/OV-SAM-6D/SAM-6D/Instance_Segmentation_Model/run_ism.sh
pravin25: done, removed the dead OV_SAM_DEVICE export/usages from run_ism.sh; OV_DEVICE is now the only device variable passed through, matching eval_ism_ov_bop.py / infer_ism_ov.py.

third-party-programs.txt
pravin25: done, fixed both broken cross-references in third-party-programs.txt by changing “entry above” to “entry below” for the DINOv2 (segment-anything) and FastSAM (ultralytics) entries.

modules/3d/OV-SAM-6D/SAM-6D/Pose_Estimation_Model/model/ov_pointnet2_op/ball_query.cl
pravin25: done, added the MIT PointNet2 attribution line to the OpenCL kernel ports as well (ball_query.cl, furthest_point_sampling.cl, grouping_operation.cl, gather_operation.cl) to match the corresponding C++ wrappers.

modules/3d/OV-SAM-6D/SAM-6D/Data/Example/gt_pose.json
pravin25: done, added an explicit attribution/license section in OV_ README.md for the committed Data/Example derivatives (gt_pose.json, models_info.json, mask_visib/000003_000001.png), including LM-O source URLs and the LM-O license (CC BY-SA 4.0), plus redistribution/compliance note.

modules/3d/OV-SAM-6D/SAM-6D/Pose_Estimation_Model/common_infer_utils.py
pravin25: done

third-party-programs.txt
pravin25: done

pravin25:
above addressed questions answered 

1.  done ,  this follow-up is completed. Both run_inference_custom_openvino.py and pem_model_convert_ov_ir.py now resolve libopenvino_operation_extension.so against BASE_DIR and guard with os.path.isfile + FileNotFoundError before core.add_extension.  

2.  done, RGB vs BGR — can you confirm the intended channel order? In the OpenVINO path templates end up as RGB (cv2.imread + flip) while the query crop ends up as BGR (load_im + flip), which looks inconsistent between the two branches.

3.  done, validate_gpu_ops_xml.py is a structural guard that enforces format="any" on multi-input custom ops in pem_gpu_ops.xml to avoid GPU reorder naming collisions; we plan to keep it as a CI/release check to prevent regressions. 

4.  done, confirmed race-free now. We reintroduced the local-memory barrier immediately after reading s_idx[0] in each FPS iteration, 

5.  done,  updated the Eigen entry in third-party-programs.txt to MPL-2.0, and added/kept an explicit LM-O source and license attribution note for the committed Data/Example derivatives in OV_README.md (dataset/source URLs + CC BY-SA 4.0 + redistribution note).  

6.  (a) FastSAM / ultralytics (AGPL-3.0): Optional, non-default path. Default is SAM (Apache-2.0); FastSAM runs only when the user explicitly passes --segmentor_model fastsam. No AGPL code is vendored or committed — it's a user-installed optional dependency, disclosed in third-party-programs.txt.
   (b) MAE backbone weights (CC-BY-NC 4.0): Not committed to the repo. Downloaded at setup time by download_sam6d-pem.py from the upstream SAM-6D release. I added an explicit CC-BY-NC 4.0 (non-commercial) note in OV_README.md.

allnes (approved):

allnes:
@ValentinaKats could you review the licensing side of this PR? One point needs a maintainer/legal call and can't be fixed in code.

The module is a port of SAM-6D (github.com/JiehongLin/SAM-6D), whose upstream repo has no license at the pinned commit 1c2543b (GitHub reports "None" = all rights reserved). This PR clones and patches that upstream tree (setup_from_original.sh, ism.patch/pem.patch), while openvino_contrib ships under Apache-2.0 — and an Apache-2.0 sign-off doesn't grant the right to relicense unlicensed third-party code. We'd need either written permission from the SAM-6D authors or an internal legal clearance recorded here.

Also worth a look: the default env pulls in copyleft deps — ultralytics (AGPL-3.0), blenderproc (GPL-3.0), plyfile (GPLv3+); disclosed in third-party-programs.txt, sources not committed, but the AGPL/GPL footprint is a call to make.

RoboticsYY:
Hi @pravin25 @allnes since SAM6D is a key model for our robotic customer, I'd love to know the status of this PR. Can we merge it, or are there any pending reviews/blockers?

modules/3d/OV-SAM-6D/OV_README.md
RoboticsYY: I saw OpenVINO has already released [2026.3](https://github.com/openvinotoolkit/openvino/releases/tag/2026.3.0), can customer use that tag for SAM6D?

pravin25:
> Hi @pravin25 @allnes since SAM6D is a key model for our robotic customer, I'd love to know the status of this PR. Can we merge it, or are there any pending reviews/blockers?

currently blocking this PR due to Licensing and  legal call. earlier @allnes  already mentioned  above in comment.

RoboticsYY:
> > Hi @pravin25 @allnes since SAM6D is a key model for our robotic customer, I'd love to know the status of this PR. Can we merge it, or are there any pending reviews/blockers?
> 
> currently blocking this PR due to Licensing and legal call. earlier @allnes already mentioned above in comment.

Hi @pravin25, How long is this legal procedure expected to take?

allnes:
@pravin25 please resolve conflict

pravin25:
solved merge conflicts

modules/3d/OV-SAM-6D/revert_setup.sh
```diff
@@ -0,0 +1,112 @@
...
+# Reset module-root README.md to base state (attribution only)
+cat > "$SCRIPT_DIR/README.md" <<'EOF'
+<!--
+Copyright (C) 2018-2026 Intel Corporation
+SPDX-License-Identifier: Apache-2.0
+-->
+
+# OV-SAM-6D
+
+This directory is an OpenVINO port of [SAM-6D](https://github.com/JiehongLin/SAM-6D) based on commit `1c2543b`.
```
ValentinaKats, line 100: This component doesn't have a license. It means that we don't have a right to use, copy or modify.

modules/3d/OV-SAM-6D/SAM-6D/Instance_Segmentation_Model/run_ism.sh
```diff
@@ -0,0 +1,205 @@
...
+set -euo pipefail
+
+# Run OpenVINO ISM evaluation on BOP LM-O (10 images) with setup checks.
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+DATA_ROOT="${DATA_ROOT:-$SCRIPT_DIR/../Data/BOP}"
+DATASET="${DATASET:-lmo}"
+CONDA_ENV="${CONDA_ENV:-ov_sam6d}"
+OV_DEVICE="${OV_DEVICE:-GPU}"
+SEGMENTOR_MODEL="${SEGMENTOR_MODEL:-fastsam}"
```
ValentinaKats, line 14: This model is licensed under AGPL-3.0, which conflicts with Apache 2.0 license of this repository.

ValentinaKats (changes requested):

PR #1069, opened 2026-06-10
Add openvino_hggd module: HGGD OpenVINO iGPU enablement

deepaks2:
This pull request adds a new module, openvino_hggd, to the contrib repository, providing OpenVINO support for the HGGD (Hybrid Grasp Detection and Generation) model, a state-of-the-art Efficient Heatmap-Guided 6-Dof Grasp Detection in Cluttered Scenes. The changes include documentation updates, usage guides, code for HGGD (Hybrid Grasp Detection and Generation) with OpenVINO custom extensions, and supporting files for building and running the module.

Major additions and updates:

1. New HGGD (Hybrid Grasp Detection and Generation) module integration

Added the openvino_hggd module
Implemented OpenVino extensions for pointcloud, written optimized kernels for farthest point sampling, knn and ball query.

2. Documentation and Usage Guides

3. Added EXPORT_AND_INFERENCE_GUIDE.md with step-by-step instructions for exporting HGGD models to OpenVINO, building required OpenVINO extensions, and running inference/evaluation.

Added a .gitignore to the module directory to exclude build artifacts, logs, and virtual environments.

AI Usage: Yes, reviewed and edited manually


third-party-programs.txt
```diff
@@ -5418,6 +5418,453 @@ SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
...
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with cvxopt. If not, see <https://www.gnu.org/licenses/>.
+
+-------------------------------------------------------------
+
+grasp-nms
```
allnes, line 5762: This entry asserts MIT for `grasp-nms`, but the PyPI metadata for `grasp-nms==1.0.2` (the exact pin in `setup.sh`) declares no license at all — the `license` field is empty, there are no license classifiers, and there's no LICENSE in the source repo either.

A package published with no license is "all rights reserved" by default: not redistributable and not Apache-2.0 compatible. Listing it as MIT here is unverified provenance. Please confirm the real license with the package author and cite the source; if it genuinely has no license, this needs to be resolved before the module depends on it.

third-party-programs.txt
```diff
@@ -5418,6 +5418,453 @@ SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
...
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+
+-------------------------------------------------------------
+
+dexnet (Berkeley AUTOLAB)
```
allnes: `dexnet` ships under the UC Berkeley research/non-commercial license — it grants use for "educational, research, and not-for-profit purposes" only and points to Berkeley OTL for commercial licensing. That is not compatible with Apache-2.0, which grants commercial use and redistribution.

Two things:

- It's not in `setup.sh`'s install list and no committed code imports it, so it's unclear why it's listed at all — if the module doesn't depend on dexnet, drop the entry.
- If the upstream pipeline copied from HGGD pulls it in at runtime, this needs legal review before the module lands: a non-commercial dependency in an Apache-2.0 module is a real conflict.

modules/openvino_hggd/setup.sh
```diff
@@ -0,0 +1,120 @@
...
+    scikit-image==0.25.2 \
+    matplotlib==3.10.8 \
+    pandas==2.3.3 \
+    tensorboardX==2.6.4 \
+    torchsummary==1.5.1 \
+    tqdm==4.67.3 \
+    transforms3d==0.4.2 \
+    trimesh==4.11.4 \
+    autolab_core==1.1.1 \
+    cvxopt==1.3.3 \
```
allnes: `cvxopt` is GPLv3-or-later (confirmed against its own LICENSE). It's installed into the module's runtime env here, though no committed code imports it — it comes in transitively via the graspnet evaluation path.

GPL is copyleft. A user running `pip install` locally isn't redistribution by this repo, but requiring a GPLv3 dependency for an Apache-2.0 module is worth flagging. If the shipped inference/eval path doesn't actually need `cvxopt`, drop it; otherwise please escalate so maintainers can decide.

modules/openvino_hggd/ov_gpu_extensions/kernels/knn_points_tiled.cl
```diff
@@ -0,0 +1,137 @@
...
+ * Copyright (C) 2018-2026 Intel Corporation
+ * SPDX-License-Identifier: Apache-2.0
+ */
+/*
+ * KNN Points - Tiled GPU kernel with local memory optimization
+ *
+ * Optimizations vs v2:
+ *   1. Tile p2 (source points) into __local memory for cooperative loading
+ *      → reduces global memory reads by TILE_SIZE× (256×)
+ *   2. Max-tracking MinK (matches pytorch3d CUDA approach)
```
allnes: This kernel carries only the Intel/Apache header, but its own comments describe it as a port of pytorch3d CUDA code — "matches pytorch3d CUDA approach" here and "matches pytorch3d MinK::add" at line 59; `fps_single_optimized.cl` says "matches the pytorch3d CUDA approach exactly". pytorch3d is BSD-3 (`Copyright (c) Meta Platforms, Inc. and affiliates`).

Please clarify: are these independent reimplementations of the published MinK/FPS algorithm (algorithms aren't copyrightable, so a clean-room reimpl is fine), or derived from pytorch3d's `.cu` sources? If derived, BSD-3 requires retaining Meta's copyright and license notice — add it to the affected `.cl` files (and to the CPU KNN/FPS in `pointcloud_ops.cpp`), the way other modules here attribute upstream kernels.

modules/openvino_hggd/ov_gpu_extensions/pointcloud_ops_gpu_v2.xml
```diff
@@ -0,0 +1,53 @@
+<!-- OpenVINO GPU Custom Layer Configuration for Point Cloud Operations -->
```
allnes, line 6: This file has no Apache/SPDX header. `.licenserc.yaml` applies the header check to `'**'` and only exempts `'**/*.json'`, `LICENSE`, `third-party-programs.txt` and a few third-party module trees — `.xml` under `openvino_hggd` isn't exempt, so the Copyright workflow (`skywalking-eyes`) will fail on this file. Add the standard header as an XML comment:

```xml
<!--
Copyright (C) 2018-2026 Intel Corporation
SPDX-License-Identifier: Apache-2.0
-->
```

modules/openvino_hggd/ov_gpu_extensions/ov_shim_gpu.py
```diff
@@ -0,0 +1,286 @@
...
+except ImportError:
+    from pointcloud_ops_native_gpu import create_ops, NativePointCloudOpsGPU
+
+
+_ops = None
+
+
+def get_ops() -> NativePointCloudOpsGPU:
+    global _ops
+    if _ops is None:
```
allnes, line 36: Default path uses `precision='f16'`, and the single-tensor ops pack neighbor indices as `float` into the output. FP16 represents integers exactly only up to 2048; above that there are gaps.

Realistic HGGD point counts (N2 up to thousands, `all_points_num=25600`) exceed 2048, so KNN/BallQuery indices and the FPS index column can be rounded to a wrong neighbor when the GPU runs the kernel in f16 — a silent accuracy loss that then feeds `gather`. Distances degrading in f16 is fine, but indices must stay exact. Either force the index-carrying computation to f32 (keep `INFERENCE_PRECISION_HINT=f32` for these layers, or emit indices via an i32 output), or document that f16 is only valid for N ≤ 2048. Can you confirm which precision actually reaches these custom layers at runtime?

modules/openvino_hggd/ov_gpu_extensions/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+target_link_libraries(${TARGET_NAME} PRIVATE openvino::runtime)
+
+# Compiler flags
+if(CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang")
+    target_compile_options(${TARGET_NAME} PRIVATE -O3 -march=native -ffast-math)
+endif()
+
+# Install targets
+install(TARGETS ${TARGET_NAME} LIBRARY DESTINATION lib RUNTIME DESTINATION lib)
+install(FILES
```
allnes: None of these five files exist in the tree. The real names are `pointcloud_ops_gpu_v2.xml`, `knn_points_tiled.cl`, `ball_query_single_v2.cl`, `fps_single_optimized.cl`, `fps_with_lengths_optimized.cl`, and there is no `gather.cl` at all (gather is done in numpy). `cmake --install` / `make install` errors on the missing files.

The documented `setup.sh` path only runs `cmake .. && make`, so it never hits install and the module still works (the `.so` loads from `build/`, the xml from the source dir) — which is why this slipped through. Fix the install list to the real filenames, or drop it if install isn't used.

modules/openvino_hggd/ov_gpu_extensions/pointcloud_ops.cpp
```diff
@@ -0,0 +1,953 @@
...
+            // Compute distances to all points
+            for (int64_t j = 0; j < N2; ++j) {
+                const float dx = p2_b[j * 3 + 0] - qx;
+                const float dy = p2_b[j * 3 + 1] - qy;
+                const float dz = p2_b[j * 3 + 2] - qz;
+                dist_idx[j] = {dx*dx + dy*dy + dz*dz, static_cast<int32_t>(j)};
+            }
+
+            // Partial sort to get K smallest
+            std::partial_sort(dist_idx.begin(), dist_idx.begin() + K, dist_idx.end());
```
allnes: No guard that `N2 >= K`. `dist_idx` has `N2` elements; `partial_sort` with `begin() + K` when `K > N2` walks past `end()` (UB), and the output loop then reads `dist_idx[k]` for `k` in `[0, K)` out of bounds. Same in `KNNPointsSingle::evaluate`. `validate_and_infer_types` doesn't check it.

In this module's pipeline N2 is large so it won't trigger, but these ops are registered as a general extension and can be reused by other models with small source clouds. Add a `NODE_VALIDATION_CHECK` (or clamp K to N2) in `validate_and_infer_types`.

modules/openvino_hggd/ov_gpu_extensions/kernels/fps_single_optimized.cl
```diff
@@ -0,0 +1,132 @@
...
+ *
+ * WorkSize: global=(256, B, 1) local=(256, 1, 1)
+ *
+ * Local memory: MAX_N floats (16KB) + reduction arrays (2KB) = ~18KB
+ */
+
+#pragma OPENCL EXTENSION cl_khr_fp16 : enable
+
+#define WORKGROUP_SIZE 256
+#define MAX_N 8192
```
allnes: `min_dist` lives in `__local` memory sized `MAX_N=8192`, and the loop clamps to `n_eff = min(N, MAX_N)`. If a padded cloud exceeds 8192 points (`fps` rounds N up to the next power of two ≥1024, and `all_points_num=25600` → `N_pad` up to 32768), FPS silently samples only the first 8192 points with no error — a wrong result that looks successful. The CPU `FPSSingle::evaluate` has no such cap, so CPU and GPU diverge for large N.

Either assert/return an error when `N > MAX_N`, or keep `min_dist` in global memory. What's the max N that actually reaches FPS in the pipeline?

modules/openvino_hggd/ov_gpu_extensions/pointcloud_ops_native_gpu.py
```diff
@@ -0,0 +1,290 @@
...
+            
+        Returns:
+            gathered: [B, K, D] or [B, K1, K2, D]
+        """
+        B = features.shape[0]
+        idx_shape = idx.shape
+        D = features.shape[-1]
+        
+        # Handle -1 indices (invalid)
+        valid_idx = np.maximum(idx, 0)
```
allnes: pytorch3d's `masked_gather` zeroes any `idx == -1` slot in the output. Here invalid (`-1`) indices are clamped to 0 and then `features[b, 0]` is gathered — empty ball-query slots return the first point's features instead of zeros. The C++ `MaskedGather::evaluate` does the correct zeroing, but the GPU path uses this numpy gather, not that op.

For `ball_query` with `return_nn=True` this is a silent behavioral difference from the pytorch3d op the shim replaces, and it can shift accuracy. Zero out positions where `idx < 0` after gathering.

modules/openvino_hggd/ov_gpu_extensions/kernels/fps_single_v2.cl
```diff
@@ -0,0 +1,134 @@
...
+ * 
+ * Algorithm: For each iteration k, each work item processes N/256 points,
+ * finds local max, then workgroup reduction finds global max.
+ */
+
+#pragma OPENCL EXTENSION cl_khr_fp16 : enable
+
+#define WORKGROUP_SIZE 256
+
+__kernel void fps_single_v2_kernel(
```
allnes: This kernel (and `fps_with_lengths.cl`, `knn_points_single_v2.cl`) isn't referenced by `pointcloud_ops_gpu_v2.xml` or any code — the XML uses the `_tiled` and `_optimized` variants. Dead files add maintenance surface, and `fps_single_v2`/`fps_with_lengths` read the sample index back from an FP16 output, which is exactly the precision trap the optimized variants avoid. Drop them unless they're intentionally kept as references.

modules/openvino_hggd/ov_gpu_extensions/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+set(TARGET_NAME "openvino_pointcloud_extension")
+add_library(${TARGET_NAME} MODULE 
+    pointcloud_ops.cpp 
+    ov_extension.cpp
+)
+target_link_libraries(${TARGET_NAME} PRIVATE openvino::runtime)
+
+# Compiler flags
+if(CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang")
+    target_compile_options(${TARGET_NAME} PRIVATE -O3 -march=native -ffast-math)
```
allnes: `-march=native` bakes the build machine's ISA into a shared extension and breaks if the `.so` is moved to another CPU; `-ffast-math` changes FP semantics (NaN/Inf handling), which is risky for the distance comparisons here. For a portable extension prefer a conservative baseline (`-O2`/`-O3` without `-march=native`) and drop `-ffast-math` unless you've verified it doesn't change results.

modules/openvino_hggd/export_models.py
```diff
@@ -0,0 +1,273 @@
+# Copyright (C) 2018-2026 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+"""Export HGGD AnchorNet and LocalNet to OpenVINO IR for Intel GPU.
+
+Loads PyTorch checkpoint, exports via ONNX, converts to OpenVINO IR (FP16),
+and validates on GPU.
+
+Usage:
+    conda activate hggd_xpu
```
allnes: Env name is inconsistent: `setup.sh` and `README.md` use `hggd_intel`, but this docstring (and the `setup.sh` comments) say `hggd_xpu`. Pick one so the copy-paste instructions work.

allnes (commented):
Code review of the new `openvino_hggd` module (Python/ML + C++/OpenCL custom extension for OpenVINO GPU) plus the `third-party-programs.txt` additions. Read the affected files in full; didn't build (needs an Intel GPU, a `torch+xpu` conda env, the built `.so`, and the copied upstream GraspNet dataset).

Verdict: changes needed, with a few items for maintainer/legal review. The blockers are on the licensing side, not the code:

- `grasp-nms` is listed as MIT but the pinned package declares no license (see inline).
- `dexnet` is a UC Berkeley non-commercial license, incompatible with Apache-2.0 (see inline).
- `cvxopt` is GPLv3 (see inline).

Two items without a line to attach to:

- The module isn't listed in the root `README.md` where every other module is — please add an entry. The PR description also mentions an `EXPORT_AND_INFERENCE_GUIDE.md`, but there's no such file in the diff; either add it or drop the claim.
- The feature commit has no `Signed-off-by`. If this repo enforces DCO, rebase with `git commit -s` so the check passes.

No tests are wired into CI (no job, no `labeler.yml`/`CODEOWNERS` entry), and `export_models.py` prints `max_diff(PT vs OV)` without asserting a threshold — worth adding a numeric gate. Detailed findings inline.

third-party-programs.txt
deepaks2: I downloaded grasp-nms 1.0.2 it has LICESE file with MIT in it. Please check this link. https://files.pythonhosted.org/packages/03/cf/c6d6de5a8341f71cd918f54be828778570ed775cc1e2efa16575946e009e/grasp_nms-1.0.2.tar.gz

third-party-programs.txt
deepaks2: Fixed

modules/openvino_hggd/setup.sh
deepaks2: fixed and tested

modules/openvino_hggd/ov_gpu_extensions/kernels/knn_points_tiled.cl
deepaks2: fixed and tested

modules/openvino_hggd/ov_gpu_extensions/pointcloud_ops_gpu_v2.xml
deepaks2: fixed and tested

modules/openvino_hggd/export_models.py
deepaks2: fixed and tested

modules/openvino_hggd/ov_gpu_extensions/CMakeLists.txt
deepaks2: fixed and tested

modules/openvino_hggd/ov_gpu_extensions/kernels/fps_single_v2.cl
deepaks2: fixed and tested

deepaks2: fixed and tested

modules/openvino_hggd/ov_gpu_extensions/pointcloud_ops_native_gpu.py
deepaks2: fixed and tested

modules/openvino_hggd/ov_gpu_extensions/kernels/fps_single_optimized.cl
deepaks2: fixed and tested

modules/openvino_hggd/ov_gpu_extensions/CMakeLists.txt
deepaks2: fixed and tested

modules/openvino_hggd/ov_gpu_extensions/pointcloud_ops.cpp
deepaks2: fixed and tested

modules/openvino_hggd/ov_gpu_extensions/ov_shim_gpu.py
deepaks2: fixed and tested

modules/openvino_hggd/setup.sh
```diff
@@ -0,0 +1,121 @@
...
```
allnes: cvxopt is GPL-3.0, a strong copyleft license, and setup.sh pins it as a runtime dependency of the hggd_intel environment. Shipping a module whose documented runtime pulls in GPL-3.0 needs maintainer/legal review for an Apache-2.0 project. Please confirm whether cvxopt is actually imported at runtime for this pipeline or only comes in transitively via graspnetAPI/grasp-nms; if it's not a direct requirement, drop the explicit pin, and if it is required, flag the GPL interaction for review before merge.

third-party-programs.txt
allnes: This dexnet entry is a UC Berkeley non-commercial license (use limited to educational/research/not-for-profit; commercial use requires a signed agreement). Nothing in this module imports dexnet and setup.sh doesn't install it (only autolab_core), so either the entry is spurious and should be removed, or dexnet is a real transitive dependency, in which case a non-commercial license attached to a component of an Apache-2.0 distribution needs to be escalated rather than silently listed. Please clarify which, and don't ship a non-commercial dependency as a hard requirement.

modules/openvino_hggd/ov_gpu_extensions/kernels/ball_query_single_v2.cl
```diff
@@ -0,0 +1,91 @@
+/*
+ * Copyright (C) 2018-2026 Intel Corporation
```
allnes, line 2: This kernel is one of the four actually installed and referenced by pointcloud_ops_gpu_v2.xml, but unlike knn_points_tiled.cl, fps_single_optimized.cl and fps_with_lengths_optimized.cl it carries only the Intel/Apache header with no PyTorch3D BSD-3 attribution, even though it implements the same ball_query algorithm ported from PyTorch3D and BallQuery in pointcloud_ops.cpp does carry that attribution. Add the same BSD-3 attribution block here for consistency.

modules/openvino_hggd/ov_gpu_extensions/pointcloud_ops_native_gpu.py
```diff
@@ -0,0 +1,292 @@
...
+<layer id="1" name="lengths" type="Parameter" version="opset1"><data element_type="f32" shape="{B}"/><output><port id="0" precision="FP32" names="lengths"><dim>{B}</dim></port></output></layer>
+<layer id="2" name="fps" type="FPSWithLengths" version="hggd"><data k="{k}"/><input><port id="0"><dim>{B}</dim><dim>{N_pad}</dim><dim>3</dim></port><port id="1"><dim>{B}</dim></port></input><output><port id="2" precision="FP32"><dim>{B}</dim><dim>{k}</dim><dim>4</dim></port></output></layer>
+<layer id="3" name="result" type="Result" version="opset1"><input><port id="0"><dim>{B}</dim><dim>{k}</dim><dim>4</dim></port></input></layer>
+</layers><edges><edge from-layer="0" from-port="0" to-layer="2" to-port="0"/><edge from-layer="1" from-port="0" to-layer="2" to-port="1"/><edge from-layer="2" from-port="2" to-layer="3" to-port="0"/></edges></net>'''
+            
+            model = self._get_model(key, xml)
+            result = model({"points": points_padded, "lengths": lengths})[0]
+        else:
+            # Use original FPSSingle kernel (pad N similarly)
```
allnes: When lengths is None, fps() zero-pads points up to a power-of-two bucket and then dispatches FPSSingle, which has no notion of valid length. The padded (0,0,0) rows become real candidates in the farthest-point search: for a cloud whose points sit away from the origin, an all-zero pad row is frequently the farthest point and gets selected, and its index (>= N) is only masked afterwards by np.clip(idx, 0, N-1), silently collapsing it to a valid-but-wrong index. Either always route padded inputs through FPSWithLengths passing the true N as length, or have FPSSingle ignore rows beyond N.

modules/openvino_hggd/ov_gpu_extensions/kernels/knn_points_tiled.cl
```diff
@@ -0,0 +1,160 @@
...
+ *          p2 [B, N2, 3] - source points
+ * Output:  [B, N1, K*2] - packed [dists_0..K-1, idx_0..K-1]
+ *
+ * WorkSize: global=(N1, B, 1) local=(256, 1, 1)
+ */
+
+#pragma OPENCL EXTENSION cl_khr_fp16 : enable
+
+#define TILE_SIZE 256
+#define MAX_K 64
```
allnes, line 48: The GPU kernels store neighbors in fixed private arrays best_dists[MAX_K]/best_idx[MAX_K] with MAX_K=64, but nothing enforces K <= 64. validate_and_infer_types() only checks k <= N2 for KNN and nothing at all for BallQuery's k. The default k=32 is fine, but the op accepts arbitrary k and any K > 64 writes past the private arrays, which is out-of-bounds on GPU. Add a K <= 64 NODE_VALIDATION_CHECK in validate_and_infer_types for KNNPoints(Single) and BallQuery(Single), or size the arrays from K.

modules/openvino_hggd/ov_gpu_extensions/kernels/fps_single_optimized.cl
```diff
@@ -0,0 +1,181 @@
...
+        output[out_base + 1 * OUTPUT0_PITCHES[2]] = (OUTPUT0_TYPE)py;
+        output[out_base + 2 * OUTPUT0_PITCHES[2]] = (OUTPUT0_TYPE)pz;
+        output[out_base + 3 * OUTPUT0_PITCHES[2]] = (OUTPUT0_TYPE)0;
+    }
+    barrier(CLK_LOCAL_MEM_FENCE);
+
+    /* ── Main FPS loop ── */
+    for (int k = 1; k < K_eff; k++) {
+
+        float my_max = -1.0f;
```
allnes, line 104: The recompute FPS kernel doesn't exclude already-selected points from the candidate scan: it recomputes the min distance to selected points and takes the argmax. When the remaining unselected points are duplicates of selected ones (degenerate or zero-padded clouds), every candidate has md == 0 and the kernel can re-select an already-sampled point, emitting a duplicate index. The CPU evaluate() marks min_dists[current] = -1 so it never re-selects. This diverges from the CPU reference on degenerate inputs, and combined with the FPSSingle padding path it is reachable in practice. fps_with_lengths_optimized.cl has the same issue.

README.md
```diff
@@ -50,6 +50,7 @@ Additional build instructions are available for the following modules:
 * [**ollama_OpenVINO**](./modules/ollama_openvino)
 * [**openvino-langchain**](./modules/openvino-langchain): LangChain.js integrations for OpenVINO™
 * [**OpenVino BEVFusion**](./modules/openvino_bevfusion): Check the [INSTRUCTIONS](./modules/openvino_bevfusion/EXPORT_AND_INFERENCE_GUIDE.md) for detailed usage and build instructions.
+* [**OpenVino HGGD**](./modules/openvino_hggd): Check the [INSTRUCTIONS](./modules/openvino_hggd/README.md) for detailed usage and build instructions.
```
allnes, line 55: HGGD is added to the "Additional build instructions" section only, but not to the main module overview list at the top of the file where every other module (including openvino_bevfusion) is listed with a one-line description. Add the one-line overview entry so the module appears in the canonical list too.

modules/openvino_hggd/ov_gpu_extensions/kernels/fps_single_v2.cl
```diff
@@ -0,0 +1,134 @@
+/*
```
allnes: This kernel, along with fps_with_lengths.cl and knn_points_single_v2.cl, is neither referenced by pointcloud_ops_gpu_v2.xml nor installed by CMakeLists, so it's dead code. fps_single_v2.cl documents a MAX_N=8192 truncation that the shipped kernels were specifically rewritten to remove, so keeping it around is misleading. Remove these three files (or wire them in if they're intended). knn_points_single_v2.cl additionally lacks the PyTorch3D attribution.

modules/openvino_hggd/ov_gpu_extensions/ov_shim_gpu.py
```diff
@@ -0,0 +1,299 @@
...
+    gathered = ops.gather(pts_np, idx_np)
+    
+    return torch.from_numpy(gathered).to(points.device, dtype=points.dtype)
+
+
+# ═══════════════════════════════════════════════════════════════════════════════
+# Transform Operations (pytorch3d API compatible)
+# ═══════════════════════════════════════════════════════════════════════════════
+
+def euler_angles_to_matrix(euler: torch.Tensor, convention: str = "XYZ") -> torch.Tensor:
```
allnes, line 209: euler_angles_to_matrix and matrix_to_quaternion reproduce the pytorch3d.transforms API and the standard Shepperd quaternion-from-matrix branches. The math is standard, but since these are drop-in replacements for pytorch3d and the rest of the module attributes PyTorch3D wherever code was ported, a one-line "API-compatible with pytorch3d.transforms" note here would keep provenance consistent. Separately, the __init__.py docstring references module names that don't exist (ov_shim_v3, pointcloud_ops_native).

modules/openvino_hggd/ov_gpu_extensions/pointcloud_ops_native_gpu.py
```diff
@@ -0,0 +1,292 @@
...
+        print(f"[GPU Native] Precision hint: {precision}")
+        
+        self._models = {}
+    
+    def _get_model(self, key: str, xml: str) -> ov.CompiledModel:
+        """Get or compile model from XML."""
+        if key in self._models:
+            return self._models[key]
+        
+        model = self.core.read_model(xml.encode('utf-8'))
```
allnes, line 68: Each op builds an IR XML by string-formatting shape and attribute values and reads it via read_model on a weightless bytes blob. Is this intended over building an ov.Model programmatically from opset ops? The string path is brittle (radius/K interpolated into XML, one compiled model cached per unique (B, N, K)) and bypasses the ONNX frontend that ov_extension.cpp registers. Does inference actually exercise the ov::frontend::OpExtension paths, or only the direct OpExtension via read_model?

modules/openvino_hggd/ov_gpu_extensions/ov_extension.cpp
```diff
@@ -0,0 +1,57 @@
...
+#include <openvino/core/extension.hpp>
+#include <openvino/core/op_extension.hpp>
+#include <openvino/frontend/extension.hpp>
+
+#include "pointcloud_ops.hpp"
+
+OPENVINO_CREATE_EXTENSIONS(
+    std::vector<ov::Extension::Ptr>({
+        // KNN Points - K-nearest neighbors (multi-output)
+        std::make_shared<ov::OpExtension<HGGDExtension::KNNPoints>>(),
```
allnes: Six multi-output ops (KNNPoints, BallQuery, FPS, MaskedGather, GatherMaxPool, PointGather) are registered here and fully implemented in pointcloud_ops.cpp, but no .py or .xml uses them; only the *Single and FPSWithLengths variants are wired up. Are the multi-output ops part of the intended public surface or leftovers from development? If unused, dropping them shrinks the extension and the maintenance/attribution surface.

allnes (changes requested):

allnes:
@deepaks2 Is this PR still relevant?

deepaks2:
> @deepaks2 Is this PR still relevant?

Apologies for the delay. I was working on the demo setup for an event because of that i did not push commit with new changes. I will address the comments and push the PR

modules/openvino_hggd/setup.sh
deepaks2: I am replacing cvxopt==1.3.3 with Scipy which is BSD 3-Clause. is this Okay? Once tested I wil push the changes

third-party-programs.txt
deepaks2: dexnet is not part of the code PR. But it is part of the HGGD source clone that user has to clone to run the model (same is documented). Should we remove the license from third-party-programs.txt?

modules/openvino_hggd/ov_gpu_extensions/kernels/ball_query_single_v2.cl
deepaks2: fixed

modules/openvino_hggd/ov_gpu_extensions/pointcloud_ops_native_gpu.py
deepaks2: fixed

modules/openvino_hggd/ov_gpu_extensions/kernels/knn_points_tiled.cl
deepaks2: fixed

modules/openvino_hggd/ov_gpu_extensions/kernels/fps_single_optimized.cl
deepaks2: fixed

README.md
deepaks2: fixed

modules/openvino_hggd/ov_gpu_extensions/ov_shim_gpu.py
deepaks2: fixed

modules/openvino_hggd/ov_gpu_extensions/pointcloud_ops_native_gpu.py
deepaks2: fixed

modules/openvino_hggd/ov_gpu_extensions/ov_extension.cpp
deepaks2: fixed

modules/openvino_hggd/ov_gpu_extensions/kernels/fps_single_v2.cl
deepaks2: removed all the unused code

third-party-programs.txt
deepaks2: Removed dexnet

modules/openvino_hggd/README.md
```diff
@@ -0,0 +1,98 @@
...
+
+OpenVINO inference enablement for [HGGD](https://github.com/THU-VCLab/HGGD) (Hybrid Grasp Detection and Generation) on Intel integrated GPU.
+
+## Prerequisites: Fetch Upstream HGGD Files
+
+The `customgraspnetAPI`, `dataset`, and `models` directories are not included in this module — they are identical to the upstream repository and should be copied directly from there:
+
+```bash
+git clone https://github.com/THU-VCLab/HGGD /tmp/HGGD_upstream
+cp -r /tmp/HGGD_upstream/customgraspnetAPI path_to_openvino_hggd/
```
allnes: This copies the nested DexNet implementation whose license limits use to education, research, and not-for-profit purposes. `customgraspnetAPI/__init__.py` imports the evaluation stack, so the restricted files enter the documented runtime import path even for inference.

Remove this package from the runtime path or replace it with a minimal implementation whose provenance permits Apache-2.0 distribution and commercial use.

modules/openvino_hggd/README.md
```diff
@@ -0,0 +1,98 @@
...
+
+Useful options: `--camera realsense` (default), `--top-k 50`,
+`--proc N` (worker processes), `--result-path /custom/location.npy`.
+
+Note: CF-AP is **not comparable** to the AP / AP@0.8 / AP@0.4 reported
+in the HGGD paper (it has no friction-coefficient semantics) — treat it
+as a separate metric.
+
+To reproduce the AP / AP@0.8 / AP@0.4 as reported in the HGGD paper,
+use the official evaluation toolkit from `graspnetAPI==1.2.11`
```
allnes: The `graspnetAPI==1.2.11` sdist predates the MIT license currently present in the upstream repository, contains no top-level license file, and bundles the same restricted DexNet implementation. Recommending this release introduces unresolved provenance and commercial-use restrictions.

Do not recommend this package until the exact release provenance and rights for all bundled files have been reviewed and documented.

modules/openvino_hggd/README.md
```diff
@@ -0,0 +1,98 @@
...
+```
+
+## Export Models
+
+```bash
+cd path_to_openvino_hggd
+conda activate hggd_intel
+
+python export_models.py \
+  --checkpoint-path /path/to/HGGD_realsense_checkpoint \
```
allnes: The checkpoint and preprocessed HGGD data are required for this workflow, but no artifact-specific license or redistribution grant is identified. The referenced GraspNet dataset and models are CC BY-NC-SA and therefore cannot support a general commercial Apache-2.0 workflow.

Document the exact source, immutable version, checksum, license, and permitted uses for every required checkpoint and dataset. These artifacts need legal approval before the module can be presented as usable without non-commercial restrictions.

modules/openvino_hggd/export_models.py
```diff
@@ -0,0 +1,273 @@
...
+    parser.add_argument('--anchor-k', type=int, default=6)
+    parser.add_argument('--anchor-num', type=int, default=7)
+    parser.add_argument('--ratio', type=int, default=8)
+    args = parser.parse_args()
+
+    output_dir = Path(args.output_dir)
+    output_dir.mkdir(parents=True, exist_ok=True)
+
+    # Load checkpoint
+    checkpoint = torch.load(args.checkpoint_path, map_location='cpu', weights_only=False)
```
allnes: This explicitly enables arbitrary pickle execution for a user-supplied checkpoint. The export only needs tensor state dictionaries, so executable Python objects are not required.

Load a tensor-only format, validate every expected key and value type, and reject unexpected objects. The pinned `torch==2.6.0+xpu` is also covered by the current `weights_only` bypass advisory; upgrade to a patched release before treating `weights_only=True` as a security boundary.

modules/openvino_hggd/evaluate.py
```diff
@@ -0,0 +1,365 @@
...
+    scene_ids = list(range(args.scene_l, args.scene_r))
+    if not scene_ids:
+        raise ValueError(f'Empty scene range: [{args.scene_l}, {args.scene_r})')
+
+    worker_args = [(sid, args.scene_path, args.pred_dir, args.camera,
+                    args.top_k, args.max_width) for sid in scene_ids]
+    if args.proc > 1:
+        from multiprocessing import Pool
+        with Pool(processes=args.proc) as p:
+            results = p.map(cf_eval_scene, worker_args)
```
allnes: `Pool.map` passes each six-element tuple as one positional argument, while `cf_eval_scene` requires six positional arguments. Every evaluation with `--proc` greater than one fails with `TypeError` before processing a scene.

Use `p.starmap(cf_eval_scene, worker_args)` and add a multiprocessing test covering at least two scenes.

modules/openvino_hggd/infer.py
```diff
@@ -0,0 +1,527 @@
...
+                                      grid_size=args.grid_size,
+                                      grasp_nms=args.grid_size,
+                                      reduce='max')
+
+            if batch_idx >= 1:
+                time_anchor_post += time() - start_post
+
+            if rect_gg.size == 0:
+                logging.warning(f'Frame {batch_idx}: no 2D grasps detected, skipping')
+                continue
```
allnes: Frames with no 2D proposals do not produce a prediction file, but the evaluator unconditionally loads all 256 frame files. The first empty frame therefore aborts evaluation with `FileNotFoundError`.

Save an empty prediction array with the documented shape for every processed frame, or make the evaluator treat a missing file as an empty prediction and report it explicitly.

modules/openvino_hggd/ov_gpu_extensions/kernels/fps_single_optimized.cl
```diff
@@ -0,0 +1,185 @@
...
+        const float py = (float)points[p_base + 1 * INPUT0_PITCHES[2]];
+        const float pz = (float)points[p_base + 2 * INPUT0_PITCHES[2]];
+
+        const int out_base = batch * OUTPUT0_PITCHES[0];
+        output[out_base + 0 * OUTPUT0_PITCHES[2]] = (OUTPUT0_TYPE)px;
+        output[out_base + 1 * OUTPUT0_PITCHES[2]] = (OUTPUT0_TYPE)py;
+        output[out_base + 2 * OUTPUT0_PITCHES[2]] = (OUTPUT0_TYPE)pz;
+        output[out_base + 3 * OUTPUT0_PITCHES[2]] = (OUTPUT0_TYPE)0;
+    }
+    barrier(CLK_LOCAL_MEM_FENCE);
```
allnes: Work-item 0 writes the selected index and coordinates to `__global output`, then the remaining work-items read that output after a barrier which only orders local memory. The global values are not guaranteed to be visible, so an iteration can use stale or uninitialized indices and coordinates. The same issue occurs at `fps_single_optimized.cl:166` and `fps_with_lengths_optimized.cl:104,171`.

Add `CLK_GLOBAL_MEM_FENCE` to every barrier that publishes `output` to the work-group, or move the selected values through local memory. Keep the reduction-only barriers local.

modules/openvino_hggd/ov_gpu_extensions/pointcloud_ops.cpp
```diff
@@ -0,0 +1,451 @@
...
+    const auto& p1_shape = get_input_partial_shape(0);
+    const auto& p2_shape = get_input_partial_shape(1);
+
+    NODE_VALIDATION_CHECK(this, p1_shape.rank().is_static() && p1_shape.rank().get_length() == 3,
+                          "p1 must be 3D [B, N1, 3]");
+    NODE_VALIDATION_CHECK(this, p2_shape.rank().is_static() && p2_shape.rank().get_length() == 3,
+                          "p2 must be 3D [B, N2, 3]");
+
+    // GPU kernel uses fixed private arrays of MAX_K=64; k > 64 is OOB on GPU
+    NODE_VALIDATION_CHECK(this, m_k <= 64,
```
allnes: `k == 0` passes validation. The OpenCL kernel then evaluates `best_dists[K - 1]`, which indexes `-1` and causes an out-of-bounds access.

Require `0 < k && k <= 64` in validation and add a negative test for zero.

modules/openvino_hggd/ov_gpu_extensions/pointcloud_ops.cpp
```diff
@@ -0,0 +1,451 @@
...
+    const auto& pts_shape = inputs[0].get_shape();
+    const int64_t B = pts_shape[0];
+    const int64_t N = pts_shape[1];
+    const int64_t K = m_k;
+
+    #pragma omp parallel for
+    for (int64_t b = 0; b < B; ++b) {
+        const float* pts_b = points + b * N * 3;
+        float* out_b = combined + b * K * 4;
+        const int64_t valid_len = static_cast<int64_t>(lengths_f[b]);
```
allnes: The CPU implementation does not clamp or validate `valid_len` against the point count. A length greater than `N` makes the following loop read beyond `pts`, while the GPU and Python paths clamp the same value.

Validate that every length is finite, integral, and in `[0, N]`, or apply the same clamp on all backends. Do not silently give CPU and GPU different input contracts.

modules/openvino_hggd/ov_gpu_extensions/pointcloud_ops.cpp
```diff
@@ -0,0 +1,451 @@
...
+                                 int64_t k)
+    : Op({p1, p2}), m_k(k) {
+    constructor_validate_and_infer_types();
+}
+
+void KNNPointsSingle::validate_and_infer_types() {
+    const auto& p1_shape = get_input_partial_shape(0);
+    const auto& p2_shape = get_input_partial_shape(1);
+
+    NODE_VALIDATION_CHECK(this, p1_shape.rank().is_static() && p1_shape.rank().get_length() == 3,
```
allnes: The custom ops validate ranks but not the required coordinate dimension, compatible batch dimensions, or `f32` element types. `evaluate()` unconditionally uses three-coordinate strides and `data<float>()`; accepted shapes such as `[B,N,2]`, mismatched batches, or non-f32 tensors can cause out-of-bounds reads or reinterpret the buffer with the wrong type.

Enforce last dimension 3, compatible batch dimensions, expected element types, `N2 >= K`, and `[B]` lengths during type inference and again for dynamic runtime shapes.

modules/openvino_hggd/ov_gpu_extensions/ov_shim_gpu.py
```diff
@@ -0,0 +1,307 @@
...
+):
+    """Farthest point sampling using native GPU extension.
+    
+    Uses FPSWithLengths GPU kernel for variable-length batches - no Python loops.
+    
+    Args:
+        points: [B, N, 3] point clouds (may be zero-padded)
+        K: Number of points to sample per batch
+        lengths: [B] actual valid lengths for each batch (if provided)
+        random_start_point: Whether to start from random point (currently ignored)
```
allnes: HGGD calls PyTorch3D FPS with `random_start_point=True`, but the shim always starts from point zero. This silently changes point groups and model inputs, so the converted pipeline is not numerically equivalent to the reference implementation.

Implement the seeded PyTorch3D start-point semantics or reject this option instead of accepting and ignoring it. Add a fixed-seed comparison against the upstream operator.

modules/openvino_hggd/ov_gpu_extensions/pointcloud_ops_native_gpu.py
```diff
@@ -0,0 +1,303 @@
...
+        gpu_config = {"CONFIG_FILE": str(self.config_path)}
+        if precision == 'f16':
+            gpu_config["INFERENCE_PRECISION_HINT"] = "f16"
+        elif precision == 'f32':
+            gpu_config["INFERENCE_PRECISION_HINT"] = "f32"
+        else:
+            # Default / let GPU plugin decide
+            gpu_config["INFERENCE_PRECISION_HINT"] = "f16"
+        
+        self.core.set_property("GPU", gpu_config)
```
allnes: `--ov-device CPU` is advertised at `infer.py:177`, but importing the shim constructs this runner and hard-codes all point-cloud custom ops to `GPU`. CPU mode therefore still requires an available Intel GPU and can fail during import before argument handling completes.

Pass the selected device into the runner and provide a real CPU path, or remove `CPU` from the accepted CLI values and document the GPU requirement.

modules/openvino_hggd/README.md
```diff
@@ -0,0 +1,98 @@
...
+Useful options: `--camera realsense` (default), `--top-k 50`,
+`--proc N` (worker processes), `--result-path /custom/location.npy`.
+
+Note: CF-AP is **not comparable** to the AP / AP@0.8 / AP@0.4 reported
+in the HGGD paper (it has no friction-coefficient semantics) — treat it
+as a separate metric.
+
+To reproduce the AP / AP@0.8 / AP@0.4 as reported in the HGGD paper,
+use the official evaluation toolkit from `graspnetAPI==1.2.11`
+(`pip install --no-deps graspnetAPI==1.2.11`), which evaluates the
```
allnes: This command cannot reproduce the documented official AP evaluation. `--no-deps` omits required packages such as `cvxopt`, while release 1.2.11 contains DexNet code using the removed `np.int` alias and conflicts with the module's NumPy 2.0.2 environment.

Provide a tested, isolated, license-cleared evaluator environment with compatible pins, or remove the official-AP instructions.

modules/openvino_hggd/evaluate.py
```diff
@@ -0,0 +1,365 @@
...
+import logging
+import os
+import sys
+import xml.etree.ElementTree as ET
+
+import numpy as np
+
+# ── Collision-Free evaluation ─────────────────────────────────────────────
+# The stage 0 (grasp NMS + object assignment + top-K selection) and stage 1
+# (gripper-box collision / empty test) below are verbatim ports of the
```
allnes: This file identifies parts of the implementation as verbatim GraspNet ports, but carries only an Intel/Apache header. The third-party entry says `Copyright (c) 2019-2023 GraspNet Group`, which does not match the audited upstream license or the license-less 1.2.11 sdist.

Record the exact source URL and commit used for the port, state whether AI-assisted changes transformed any third-party code, retain the exact upstream copyright and license notice in this file, and correct the third-party entry from that source.

third-party-programs.txt
```diff
@@ -5418,6 +5418,406 @@ SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
...
+See the License for the specific language governing permissions and
+limitations under the License.
+
+-------------------------------------------------------------
+
+grasp-nms
+
+MIT License
+
+Copyright (c) 2021 graspnet contributors
```
allnes: The pinned `grasp-nms==1.0.2` source carries `Copyright (c) 2020 Gou Minghao`, not this notice. The MIT license is compatible, but attribution must be copied exactly from the distributed source.

Replace this line with the upstream copyright notice and retain the MIT text required for redistribution.

modules/openvino_hggd/setup.sh
```diff
@@ -0,0 +1,117 @@
...
+    echo "Creating conda environment: $ENV_NAME (Python 3.10)"
+    conda create -n "$ENV_NAME" python=3.10 -y
+fi
+
+echo "Installing packages into $ENV_NAME..."
+conda run -n "$ENV_NAME" pip install --upgrade pip
+
+# Core Intel XPU runtime libraries used by hggd_intel.
+conda run -n "$ENV_NAME" pip install \
+    intel-cmplr-lib-rt==2025.0.2 \
```
allnes: The environment installs `intel-cmplr-lib-rt`, `intel-cmplr-lib-ur`, `intel-cmplr-lic-rt`, and `intel-sycl-rt` under the Intel EULA, plus `tcmlib` and `umf` under custom Intel terms. These are neither Apache-2.0 dependencies nor covered by the added third-party entries.

Document them as externally installed proprietary or custom-licensed runtimes, confirm that the intended packaging does not redistribute them, and obtain legal approval before shipping an environment or container that includes them.

modules/openvino_hggd/.gitignore
```diff
@@ -0,0 +1,31 @@
+# OpenVINO exported models
```
allnes: Both this file and `modules/openvino_hggd/README.md` lack the project copyright/SPDX header, and the repository license check currently fails on exactly these paths.

Add the standard comment-form Apache-2.0 header accepted by the repository checker.

modules/openvino_hggd/ov_gpu_extensions/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+# Compiler flags
+if(CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang")
+    target_compile_options(${TARGET_NAME} PRIVATE -O3)
+endif()
+
+# Install targets
+install(TARGETS ${TARGET_NAME} LIBRARY DESTINATION lib RUNTIME DESTINATION lib)
+install(FILES
+    pointcloud_ops_gpu_v2.xml
+    kernels/knn_points_tiled.cl
```
allnes: `install(FILES ... DESTINATION lib)` flattens these kernel files into `lib`, while the extension XML refers to `kernels/<name>.cl`. An installed extension cannot resolve those paths, although source-tree execution hides the problem.

Install the kernels under `lib/kernels`, or update the XML and runtime lookup to match the installed layout. Add an install-tree smoke test.

modules/openvino_hggd/ov_gpu_extensions/ov_shim_gpu.py
```diff
@@ -0,0 +1,307 @@
...
+    dists, idx = ops.ball_query(p1_np, p2_np, k=K, radius=radius)
+    
+    dists_t = torch.from_numpy(dists).to(p1.device, dtype=p1.dtype)
+    idx_t = torch.from_numpy(idx).to(p1.device, dtype=torch.int64)
+    
+    if return_nn:
+        knn = masked_gather(p2, idx_t)
+        return dists_t, idx_t, knn
+    
+    return dists_t, idx_t
```
allnes: PyTorch3D `ball_query` returns three fields, and HGGD's non-default branch destructures three values. The shim returns two, so selecting that branch raises an unpacking error.

Match the PyTorch3D result contract, including a third `knn` field set to `None` when neighbor coordinates are not requested.

modules/openvino_hggd/export_models.py
```diff
@@ -0,0 +1,273 @@
...
+
+    # Validate on GPU
+    core = ov.Core()
+    compiled = core.compile_model(str(xml_path), ov_device)
+    ov_result = compiled(dummy_input.numpy())
+    for i, (name, pt_out) in enumerate(zip(
+        ["loc_map", "cls_mask", "theta_offset", "depth_offset", "width_offset", "features"],
+        pt_outputs,
+    )):
+        max_diff = np.abs(pt_out.numpy() - ov_result[i]).max()
```
allnes: Export validation only prints the maximum difference. It does not reject shape mismatches, NaNs, or differences above a tolerance, so a broken conversion still reports successful export.

Assert output shapes, finite values, and per-output `rtol` and `atol` thresholds; terminate with a non-zero status when validation fails.

modules/openvino_hggd/infer.py
```diff
@@ -0,0 +1,527 @@
...
+            if not os.path.exists(save_dir):
+                os.makedirs(save_dir)
+            gg.save_npy(save_path)
+
+            if batch_idx >= 1:
+                time_wall += time() - frame_start
+
+    # Time stats
+    if batch_idx >= 2:
+        n = batch_idx - 1
```
allnes: Timing starts at `batch_idx >= 1`, so after `F` frames the number of measured frames is `F - 1`, equal to the final `batch_idx`, not `batch_idx - 1`. The reported averages are inflated, and skipped frames make the denominator diverge further.

Maintain an explicit counter incremented only when a timed sample is added.

modules/openvino_hggd/ov_gpu_extensions/kernels/fps_single_optimized.cl
```diff
@@ -0,0 +1,185 @@
...
+
+    /* ── Main FPS loop ── */
+    for (int k = 1; k < K_eff; k++) {
```
allnes: Each point recomputes its distance to every previously selected point on every FPS iteration. This makes the kernel `O(N*K²)` instead of the usual `O(N*K)` minimum-distance update; the documented `K=512` path performs billions of distance calculations for normal HGGD batches.

Store and update each point's running minimum distance, or provide device benchmarks demonstrating that this implementation meets the required latency. Include numeric comparisons against PyTorch3D for random starts, zero lengths, `K > N`, and duplicate points.

modules/openvino_hggd/README.md
```diff
@@ -0,0 +1,98 @@
+# HGGD — OpenVINO iGPU Enablement
+
+OpenVINO inference enablement for [HGGD](https://github.com/THU-VCLab/HGGD) (Hybrid Grasp Detection and Generation) on Intel integrated GPU.
+
+## Prerequisites: Fetch Upstream HGGD Files
+
+The `customgraspnetAPI`, `dataset`, and `models` directories are not included in this module — they are identical to the upstream repository and should be copied directly from there:
+
+```bash
+git clone https://github.com/THU-VCLab/HGGD /tmp/HGGD_upstream
```
allnes: Copying code from the moving default branch makes both behavior and provenance depend on the day setup is run. A later upstream change can also invalidate the reviewed license and attribution.

Check out an audited commit or immutable tag before copying files and record that revision in the third-party notice.

modules/openvino_hggd/README.md
```diff
@@ -0,0 +1,103 @@
...
+cp -r /tmp/HGGD_upstream/customgraspnetAPI path_to_openvino_hggd/
```
allnes, line 16: This copies the complete `customgraspnetAPI` tree, including Dex-Net code whose license permits only educational, research, and not-for-profit use. That restriction is incompatible with presenting the resulting module as unrestricted Apache-2.0 software. Copy only the required permissively licensed files, or obtain and document commercial-compatible permission; pin the exact HGGD revision and retain its notices.

third-party-programs.txt
```diff
@@ -5418,6 +5418,406 @@ SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
...
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+
+-------------------------------------------------------------
+
+graspnetAPI
+
+MIT License
+
+Copyright (c) 2019-2023 GraspNet Group
```
allnes, line 5453: This MIT notice is not traceable to `graspnetAPI==1.2.11`. Its 2023 sdist contains no license file or license metadata, and the repository added an MIT license only in 2025 with `Copyright (c) 2025 GraspNet`. Pin a revision carrying a valid grant, confirm that it covers the ported code, and reproduce the actual notice instead of the unsupported 2019-2023 attribution.

modules/openvino_hggd/README.md
```diff
@@ -0,0 +1,103 @@
...
+## Run
+
+```bash
+cd path_to_openvino_hggd
+conda activate hggd_intel
+
+bash run.sh \
+  /path/to/HGGD_realsense_checkpoint \
+  /path/to/dataset/6dto2drefine_realsense \
+  /path/to/graspnet \
```
allnes, line 50: The documented workflow requires GraspNet data and an HGGD checkpoint, but does not state their terms. GraspNet's official site restricts its data, labels, code, and models to non-commercial use under CC BY-NC-SA, while no explicit license was found for the referenced HGGD checkpoint. Document exact artifact URLs and terms and obtain legal approval before treating this as a generally usable Apache-2.0 module.

modules/openvino_hggd/setup.sh
```diff
@@ -0,0 +1,117 @@
...
+conda run -n "$ENV_NAME" pip install \
+    torch==2.6.0+xpu \
+    torchvision==0.21.0+xpu \
+    pytorch-triton-xpu==3.2.0 \
+    --index-url https://download.pytorch.org/whl/xpu
+
+conda run -n "$ENV_NAME" pip install \
+    openvino==2026.1.0 \
+    numpy==2.0.2 \
+    scipy==1.15.3 \
```
allnes, line 56: The documented copy step still fetches upstream `quality.py`, which optionally imports `cvxopt` and then unconditionally accesses `cvx.solvers` during package import. `setup.sh` no longer installs `cvxopt`, so `from customgraspnetAPI import Grasp` fails before inference starts. The SciPy rewrite mentioned in the commit history is not part of this PR. Include that patch in a pinned source tree or remove the import path.

modules/openvino_hggd/ov_gpu_extensions/kernels/fps_single_optimized.cl
allnes, line 99: Work-item 0 writes the selected index to global `output`, then every work-item reads it in the next iteration. A local-memory fence does not make that global write visible. Use `barrier(CLK_LOCAL_MEM_FENCE | CLK_GLOBAL_MEM_FENCE)` after initialization and after each selected-point write here and in `fps_with_lengths_optimized.cl:104,171`.

modules/openvino_hggd/infer.py
```diff
@@ -0,0 +1,527 @@
...
+    cloud = points_all[:, :3].clone()
+    mfcdetector = ModelFreeCollisionDetectorCPU(cloud, voxel_size=0.01, mode=mode)
+    no_collision_mask = mfcdetector.detect(pred_gg, approach_dist=0.05)
+    collision_free_gg = pred_gg[no_collision_mask]
+    return collision_free_gg, no_collision_mask
+
+
+# ── Argument Parser ────────────────────────────────────────────────────
+parser = argparse.ArgumentParser()
+parser.add_argument('--ov-device', required=True, choices=['CPU', 'GPU'], help='OpenVINO device')
```
allnes, line 177: `CPU` is advertised, but the shim is initialized before argument parsing and `NativePointCloudOpsGPU` always configures and compiles for `"GPU"`. CPU-only execution therefore still requires a working GPU plugin. Pass the selected device into the shim, provide CPU implementations, or remove `CPU` from the interface.

modules/openvino_hggd/ov_gpu_extensions/ov_shim_gpu.py
allnes, line 170: This changes reachable HGGD behavior: `dataset/pc_dataset_tools.py` calls FPS with `random_start_point=True`, while both kernels always start at index 0. Implement the random start and pass it to the native operation, or reject the option instead of silently changing sampling semantics.

modules/openvino_hggd/infer.py
allnes, line 334: A frame with no 2D grasps is skipped without writing an empty prediction, but `evaluate.py` unconditionally loads all 256 frame files. One empty frame therefore aborts evaluation with `FileNotFoundError`. Save an empty `GraspGroup` before continuing, or make the evaluator treat a missing frame as an empty prediction.

modules/openvino_hggd/ov_gpu_extensions/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+    kernels/ball_query_single_v2.cl
+    kernels/fps_single_optimized.cl
+    kernels/fps_with_lengths_optimized.cl
+    DESTINATION lib
```
allnes, line 49: Installation flattens the XML and all kernels into `lib`, while the installed XML still resolves `kernels/*.cl`. Build-tree execution works, but an installed extension cannot find its kernels. Preserve the `kernels/` subdirectory or update the installed XML paths.

modules/openvino_hggd/ov_gpu_extensions/pointcloud_ops_native_gpu.py
```diff
@@ -0,0 +1,303 @@
...
+from pathlib import Path
+from typing import Optional, Tuple
+import numpy as np
+import openvino as ov
+
+
+class NativePointCloudOpsGPU:
+    """OpenVINO native point cloud ops with GPU + FP16 support for performance."""
+    
+    def __init__(self, precision: str = 'f16'):
```
allnes, line 25: The public class defaults to FP16 even though the packed outputs store indices as floating-point values and this module documents that indices above 2048 become inexact. The factory's `"default"` path also forces FP16. Default both entry points to FP32, or return indices through an integer output.

modules/openvino_hggd/ov_gpu_extensions/ov_shim_gpu.py
allnes, line 152: PyTorch3D `ball_query` always returns the three-field `_KNN` result; `knn` is `None` when `return_nn=False`. The copied HGGD alternate path destructures three values, so this two-value return raises immediately. Return the same named tuple shape as PyTorch3D.

modules/openvino_hggd/export_models.py
```diff
@@ -0,0 +1,273 @@
...
+        log.info(f"  {name}: shape={pt_out.shape}, max_diff(PT vs OV@{ov_device})={max_diff:.6f}")
```
allnes, line 173: This validation cannot fail when conversion produces incorrect values. There are also no automated tests for the custom operations or kernels. Add tolerance assertions and CPU-reference tests covering non-power-of-two sizes, `K` boundaries, empty lengths, duplicate points, FP32/FP16, and repeated execution; wire at least the CPU-capable subset into CI.

modules/openvino_hggd/ov_gpu_extensions/kernels/fps_single_optimized.cl
```diff
@@ -0,0 +1,185 @@
...
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
+ * ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+/*
+ * Farthest Point Sampling - Recomputing O(N*K*K) kernel
```
allnes, line 36: The kernel rescans every previously selected point on every FPS iteration, changing the normal O(N*K) algorithm to O(N*K^2). `run.sh` uses `K=512` for up to 48 groups, so this adds 130,816 selected-point distance checks per candidate. Preserve per-point minimum distances in scratch/global storage or provide measurements showing that this cost is acceptable.

modules/openvino_hggd/evaluate.py
allnes, line 32: `evaluate.py` explicitly contains verbatim GraspNetAPI ports, and `infer.py` substantially adapts HGGD inference and collision code, but both carry only the Intel/Apache header. Record the exact source revisions and retain the applicable upstream copyright and license notices. Resolve the GraspNetAPI license grant before assigning Apache-2.0 to its port.

modules/openvino_hggd/setup.sh
allnes, line 38: Four pinned runtime packages declare the proprietary Intel End User License Agreement, and none of the seven packages in this block is recorded in `third-party-programs.txt`. They are downloaded rather than vendored, but their terms and necessity still need review, especially because Torch computation is patched to CPU and OpenVINO owns the GPU path. Document all seven and remove those that are not required.

third-party-programs.txt
allnes, line 5766: This does not match the license shipped in the `grasp-nms==1.0.2` sdist, which says `Copyright (c) 2020 Gou Minghao`. Copy the notice verbatim from the distributed package.

allnes (changes requested):

PR #1085, opened 2026-06-22
CDPN: fp16 and int8 enablement

pj3iL:
fp16 and int8 enablement (mixed precision for accuracy) for CDPN model.


.gitmodules
allnes, line 1: this empty .gitmodules (0 bytes) fails the License header check CI job (invalid: 1 → .gitmodules). No submodule is declared — OV_README.md uses a manual git clone ... cdpn_repo, not a submodule. Please delete this file. If a submodule was intended, then the manual-clone flow in OV_README.md contradicts it and one approach must be chosen.

modules/3d/CDPN/ov_export.py
```diff
@@ -159,7 +161,7 @@ def _build_trans_decode_subgraph(opset, raw_trans_4d, bbox_wh, crop_meta,
     return opset.concat([tx, ty, pred_depth], axis=3)
 
 
-def _make_CdpnPreprocess(ov, OVType):
+def _make_CdpnPreprocess():
     """Factory: return the CdpnPreprocess graph-topology wrapper class.
 
     Defined as a factory because ov/OVType are lazy-imported inside each
```
allnes: the refactor moved ov/opset/OVType to module-level imports (top of file), so they are no longer lazy-imported inside each export function. This rationale is now misleading — please update or drop it.

modules/3d/CDPN/ov_export.py
```diff
@@ -715,6 +717,1008 @@ def clone_with_new_inputs(self, new_args):
...
+        ET.SubElement(rt_info, 'attribute', {
+            'name': attr_name,
+            'version': '0',
+        })
+        marked += 1
+    tree.write(xml_path, encoding='UTF-8', xml_declaration=True)
+    return marked
+
+
+def export_nn_mixed_fp16_ir(model_path, output_dir,
```
allnes: export_nn_mixed_fp16_ir (870), export_extnn_mixed_fp16_ir (1075) and export_e2e_mixed_fp16_ir (1234) are ~80% identical: find rot entry / final conv, collect FP32 constants, ConvertFP32ToFP16, restore constants, re-cast rot entry input, wrap island, save, _inject_rt_attribute, near-identical prints. Please extract a shared core with hooks for the composite-only steps (_prepare_composite_fp16_inputs, _prepare_e2e_pose_concat_inputs). The duplication is already causing drift — see the dead-code notes below.

modules/3d/CDPN/ov_export.py
```diff
@@ -715,6 +717,1008 @@ def clone_with_new_inputs(self, new_args):
...
+    fp32_constants = {}
+    for op in original.get_ordered_ops():
+        if (op.get_friendly_name() in precise_names
+                and op.get_type_name() == 'Constant'
+                and _is_float_output(op.output(0))):
+            fp32_constants[op.get_friendly_name()] = op.get_data().astype(np.float32)
+
+    mixed_model = core.read_model(model_path)
+    preprocess_converts = _prepare_composite_fp16_inputs(mixed_model)
+    pose_concat_converts = 0
```
allnes: Dead in the EXTNN path: pose_concat_converts is set to 0 and never updated, yet it is printed at lines 1194–1196 as Inserted E2E pose concat FP32 converts: 0. This was copy-pasted from the E2E variant (line 1289) where it is actually computed. EXTNN has no pose concat — remove both the variable and its print.

modules/3d/CDPN/ov_export.py
```diff
@@ -715,6 +717,1008 @@ def clone_with_new_inputs(self, new_args):
...
+        to_f32.set_friendly_name(op.get_friendly_name() + '/input_fp32')
+        op.input(0).replace_source_output(to_f32.output(0))
+
+    fp16_inputs, fp16_outputs = _wrap_fp16_island_ops(
+        mixed_model, fp16_island_names)
+
+    precise_boundary_converts = _convert_precise_region_inputs_to_f32(
+        mixed_model, precise_names, skip_source_names=fp16_island_names)
+
+    precise_output_converts = 0
```
allnes: Dead assignment — immediately overwritten by _convert_region_outputs_to_f16(...) on line 1158. Drop line 1157.

modules/3d/CDPN/ov_export.py
```diff
@@ -715,6 +717,1008 @@ def clone_with_new_inputs(self, new_args):
...
+
+        for port, output in enumerate(op.outputs()):
+            if not _is_float_output(output):
+                continue
+            for target_input in list(output.get_target_inputs()):
+                target = target_input.get_node()
+                if (target.get_type_name() != 'Result'
+                        and target.get_friendly_name() in op_names):
+                    continue
+                to_f32 = opset.convert(output, ov.Type.f32)
```
allnes: This creates a separate Convert→f32 node for every consumer of the island output. Compare with _convert_region_outputs_to_f16 (lines 1059–1067) which correctly hoists a single Convert (if to_f16 is None). Please hoist one Convert per output port here too, to avoid redundant nodes in the IR.

modules/3d/CDPN/ov_export.py
```diff
@@ -715,6 +717,1008 @@ def clone_with_new_inputs(self, new_args):
...
+            ', '.join(missing), entry_names))
+    return entry_names
+
+
+def _find_rot_final_conv_name(model):
+    matches = []
+    for op in model.get_ordered_ops():
+        name = op.get_friendly_name()
+        if (op.get_type_name() == 'Convolution'
+                and 'rot_head_net.features.27/' in name):
```
allnes: Hard-coded op-name marker rot_head_net.features.27/. This depends on the exact torch→OV friendly-name mapping and on the RotHead layer count; a torchvision/openvino bump can silently change it. The raise on mismatch (fail-fast) is good, but please add a comment stating this is tied to the OV/torchvision versions pinned in ov_requirements.txt and ideally derive the index from the head config instead of a literal.

modules/3d/CDPN/ov_export.py
```diff
@@ -715,6 +717,1008 @@ def clone_with_new_inputs(self, new_args):
...
+def _nn_head_layer_groups(model):
+    groups = {}
+    for idx, bn_idx in enumerate((1, 4, 7)):
+        groups['trans_conv{}'.format(idx)] = _nn_head_feature_group_names(
+            model, 'trans_head_net', bn_idx)
+    groups['trans_mlp0'] = _nn_trans_linear_group_names(model, 0, 1)
+    groups['trans_mlp1'] = _nn_trans_linear_group_names(model, 2, 3)
+    groups['trans_mlp2'] = _nn_trans_linear_group_names(model, 4)
+
+    for idx, bn_idx in enumerate((1, 4, 7, 10, 13, 16, 19, 22, 25)):
```
allnes: Magic index tuples (1,4,7) / (1,4,7,...,25) and the linear markers trans_head_net.linears.{}/aten::linear/MatMul (line 1411) encode the exact op layout of the exported graph. Same fragility as note #7 — a version bump can break the mapping silently. Please document the assumption and, if feasible, compute these from the module structure.

modules/3d/CDPN/ov_export.py
```diff
@@ -732,39 +1736,87 @@ def clone_with_new_inputs(self, new_args):
                         help='Also export EXTNN model with pre/post processing')
     parser.add_argument('--e2e', action='store_true',
                         help='Also export E2E model with custom extension ops')
+    parser.add_argument('--fp16_nn', action='store_true',
+                        help='Also export NN mixed FP16 IR')
+    parser.add_argument('--fp16_extnn', action='store_true',
```
allnes, line 1795: The help says (requires --extnn), but the main logic uses if args.extnn or args.fp16_extnn: (line 1789) and auto-runs the EXTNN export when --fp16_extnn is passed alone. So --extnn is not required. Either fix the help text to reflect the auto-enable, or enforce the dependency. Same for --fp16_e2e (1743 / 1807).

modules/3d/CDPN/ov_export.py
```diff
@@ -732,39 +1736,87 @@ def clone_with_new_inputs(self, new_args):
...
 
     # Always export the NN body first
     nn_results = export_to_ov_ir(model, args.output_dir, args.basename, args.verify)
 
+    if args.fp16_nn:
+        export_nn_mixed_fp16_ir(
+            nn_results['xml'], args.output_dir,
+            basename=args.basename + '_fp16')
+
+    int8_nn_results = None
```
allnes: int8_nn_results is assigned (1778, 1780) but never read afterwards. Dead variable — either use the result or drop the assignment and call export_nn_int8_ir(...) directly.

modules/3d/CDPN/ov_export.py
```diff
@@ -732,39 +1736,87 @@ def clone_with_new_inputs(self, new_args):
...
+                        choices=('PERFORMANCE', 'MIXED'),
+                        help='NNCF quantization preset (default: PERFORMANCE)')
     parser.add_argument('--extension', type=str, default=None,
                         help='Path to cdpn_extension.so')
     args = parser.parse_args()
 
     if args.extnn and args.e2e:
         parser.error('Export --extnn and --e2e in separate invocations; '
                      'each composite export rewires the converted NN body.')
+    if ((args.int8_nn)
```
allnes: Redundant parentheses / awkward line break around a single boolean. Simplify to if args.int8_nn and not args.dataset_dir:.

modules/3d/CDPN/ov_export.py
```diff
@@ -715,6 +717,1008 @@ def clone_with_new_inputs(self, new_args):
...
+
+    quantized_model = nncf.quantize(
+        ov_model,
+        calibration_dataset,
+        subset_size=calibration_size,
+        target_device=_nncf_target_device(nncf, target_device),
+        preset=_nncf_quantization_preset(nncf, preset),
+        ignored_scope=ignored_scope)
+
+    fp16_op_names, fp16_constant_names, fp16_region_names = (
```
allnes: fp16_constant_names is unpacked but never used in export_nn_int8_ir. Use _ for the unused slot to signal intent (minor).

modules/3d/CDPN/ov_export.py
```diff
@@ -715,6 +717,1008 @@ def clone_with_new_inputs(self, new_args):
...
+        for inp in op.inputs():
+            source = inp.get_source_output().get_node()
+            if source.get_type_name() == 'Constant':
+                names.add(source.get_friendly_name())
+    return names
+
+
+def _is_float_output(output):
+    elem_type = str(output.get_element_type()).lower()
+    return ('float' in elem_type or 'f16' in elem_type
```
allnes: str(Type.f32) is 'f32' and str(Type.f16) is 'f16', so the 'float' in elem_type branch never matches and is dead; the check is also loose (substring match). Consider comparing against output.get_element_type() explicitly (in (Type.f16, Type.f32, Type.bf16)) for a precise, self-documenting test.

modules/3d/CDPN/ov_infer.py
```diff
@@ -43,7 +43,7 @@
 
     # Standalone test (E2E model with extension):
     python ov_infer.py --model checkpoints/cdpn_stage3_e2e.xml \
-        --extension ov_plugins/build/cdpn_cpu/cdpn_cpu_extension.so \
+        --extension ov_plugins/build/cdpn_cpu_extension.so \
```
allnes: The built artifact is cdpn_extension.so, not cdpn_cpu_extension.so (see CMakeLists.txt → add_library(cdpn_extension MODULE ...), build.sh, and OV_README.md). Fix to ov_plugins/build/cdpn_extension.so.

modules/3d/CDPN/ov_infer.py
```diff
@@ -130,16 +134,34 @@ def __init__(self, model_path, obj_info,
...
+
+        self.int8_quantization_points = sum(1 for op in self.model.get_ordered_ops()
+            if op.get_type_name() == 'FakeQuantize')
+        if self.int8_quantization_points:
+            print('[CdpnOVInference] Model contains {} INT8 quantization points'.format(
+                self.int8_quantization_points))
+
         self.compiled = self.core.compile_model(self.model, device, config)
 
+        self._infer_request = self.compiled.create_infer_request()
```
allnes, line 162: Good change (reusing one infer request avoids per-call allocation). Just flag for the record: self._infer_request is now shared state, so a single CdpnOVInference instance is not thread-safe for concurrent __call__s. If that's acceptable, a one-line note in the class docstring would help future users.

modules/3d/CDPN/ov_infer.py
```diff
@@ -706,8 +732,20 @@ def main():
         obj_info=obj_info,
         device=device,
         extension_path=args.extension,
+        inference_precision=args.infer_precision,
     )
 
+    if args.int8_nn:
```
allnes, line 737: The --int8_nn guard only checks that the model is nn_only and has ≥1 FakeQuantize. It doesn't validate that quantization actually covers the intended head groups. Given the mixed-precision policy in export_nn_int8_ir, consider a stronger sanity check (or at least log the quantization-point count), otherwise a mostly-FP16 model with a single FQ would pass.

modules/3d/CDPN/ov_requirements.txt
```diff
@@ -1,10 +1,14 @@
+# Copyright (C) 2018-2026 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
 # CDPN Dependencies
 torch==2.10.0
 torchvision==0.25.0 --index-url https://download.pytorch.org/whl/xpu
```
allnes: In a requirements file, --index-url is a global option (not per-line): it makes the XPU wheel index the primary index for all packages below (openvino==2026.2.0, nncf==3.2.0, numpy, …), which can break resolution or pull unexpected wheels. Use --extra-index-url on its own line, or move torch/torchvision to a separate install step.

modules/3d/CDPN/requirements.txt
```diff
@@ -1,10 +1,14 @@
...
+
 # CDPN Dependencies
 torch==2.10.0
 torchvision==0.25.0 --index-url https://download.pytorch.org/whl/xpu
-openvino==2026.1.0
 numpy==1.26.4
 opencv-python==4.11.0.86
 scipy==1.17.1
+openvino==2026.2.0
+nncf==3.2.0
```
allnes, line 9: This PR renames requirements.txt → ov_requirements.txt, but .github/dependabot.yml tracks pip under /modules/3d/CDPN. Dependabot's pip ecosystem keys on the standard requirements.txt filename; with a non-standard name it may stop opening dependency-update PRs for this module. Please confirm Dependabot still picks it up, or keep a standard-named file. (Both nncf==3.2.0 and openvino==2026.2.0 do exist on PyPI — verified.)

modules/3d/CDPN/OV_README.md
```diff
@@ -0,0 +1,112 @@
...
+-->
+
+# Note
+
+Run the following commands to setup the CDPN repo with the required changes for OpenVINO inference.
+
+
+```bash
+# in openvino_contrib/modules/3d/CDPN/
+git clone https://github.com/LZGMatrix/CDPN_ICCV2019_ZhigangLi cdpn_repo
```
allnes, line 33: The removed README.md carried upstream attribution: the ICCV 2019 paper citation (@InProceedings{Li_2019_ICCV...}), authors (Zhigang Li, Gu Wang, Xiangyang Ji) and the Tsinghua acknowledgement. Upstream is Apache-2.0, so runtime cloning is fine, but in the spirit of Apache-2.0 §4 please restore a short attribution block here (fork of ...@625f9a8, Apache-2.0 + the BibTeX citation).

modules/3d/CDPN/OV_README.md
```diff
@@ -0,0 +1,112 @@
...
+cd cdpn_repo
+git checkout 625f9a8
+patch -p1 < ../cdpn_changes.patch
```
allnes: The setup isn't idempotent: git clone fails if cdpn_repo exists and patch -p1 fails/double-applies on a rerun. The whole flow also hinges on an external clone pinned at 625f9a8 — if the repo/commit disappears, it can't be reproduced. Consider shipping an idempotent setup.sh/Makefile (skip clone if present, patch --forward/check-if-applied) and documenting the upstream-availability dependency.

modules/3d/CDPN/OV_README.md
```diff
@@ -0,0 +1,112 @@
...
+grep -v -E '^\s*(#|$)' ../copy_files_to_cdpn_repo.txt | xargs -I {} sh -c 'mkdir -p "$(dirname "./{}")" && cp "../{}" "./{}"'
+pip install --force-reinstall -r ov_requirements.txt
```
allnes: --force-reinstall reinstalls everything including torch/torchvision — heavy and rarely needed. pip install -r ov_requirements.txt (optionally --upgrade) is enough.

modules/3d/CDPN/OV_README.md
```diff
@@ -0,0 +1,112 @@
...
+Run the following commands to setup the CDPN repo with the required changes for OpenVINO inference.
+
+
+```bash
+# in openvino_contrib/modules/3d/CDPN/
+git clone https://github.com/LZGMatrix/CDPN_ICCV2019_ZhigangLi cdpn_repo
+cd cdpn_repo
+git checkout 625f9a8
+patch -p1 < ../cdpn_changes.patch
+grep -v -E '^\s*(#|$)' ../copy_files_to_cdpn_repo.txt | xargs -I {} sh -c 'mkdir -p "$(dirname "./{}")" && cp "../{}" "./{}"'
```
allnes: Two portability concerns: (1) \s in grep -E is a GNU extension — BSD/macOS grep won't match it, silently keeping comment/blank lines. Use grep -v -E '^[[:space:]]*(#|$)'. (2) Injecting {} into sh -c '...' is brittle (whitespace/quoting); prefer xargs -I{} cp --parents ... or a small loop. Fine for a controlled Linux workspace, but worth hardening.

modules/3d/CDPN/cdpn_changes.patch
```diff
@@ -0,0 +1,75 @@
+# Copyright (C) 2018-2026 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+--- /tmp/cdpn_diff_work/cdpn_repo/.gitignore	2026-06-29 04:59:45.384116198 +0000
```
allnes: The patch headers leak the author's local build path (/tmp/cdpn_diff_work/cdpn_repo/...) and the timestamps are inverted (the --- "old" side is newer, 06-29, than the +++ "new" side, 06-11). It applies with -p1 since the +++ side is CDPN/..., but the headers are confusing. Please regenerate with clean relative paths (e.g. diff -u a/ b/) so the patch is self-explanatory and reproducible.

modules/3d/CDPN/cdpn_changes.patch
```diff
@@ -0,0 +1,75 @@
...
++++ CDPN/.gitignore	2026-06-11 08:15:00.277754244 +0000
```
allnes: The +++ paths use a CDPN/ prefix while OV_README.md clones into cdpn_repo/ and applies with patch -p1 (stripping one component → .gitignore). This works only because every hunk targets a top-level or lib/... path where -p1 happens to line up. It's implicit and easy to break — please make the strip level and target prefix consistent and documented.

modules/3d/CDPN/.gitignore
```diff
@@ -1,136 +1,3 @@
...
-*$py.class
-
-# C extensions
-*.so
-
-# Distribution / packaging
-.Python
+# Copyright (C) 2018-2026 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
 build/
```
allnes, line 3: OV_README.md clones into cdpn_repo/, but only build/ is ignored — the cloned upstream tree will appear as untracked files in the module. Add cdpn_repo/ here.

allnes:
https://github.com/openvinotoolkit/openvino_contrib/blob/b5e18c1ccc884c028d182712c3960ec03fcb22c1/.licenserc.yaml#L28

After this PR the vendored third-party fork is removed; everything left under modules/3d/CDPN/ is Intel-authored with proper SPDX headers (verified). This blanket ignore now hides Intel files from the license check — which is exactly how the empty .gitmodules slipped through. Narrow it to modules/3d/CDPN/cdpn_repo/** (the runtime clone) so Intel sources are checked normally. Update the accompanying comment on line 27 too.

modules/3d/CDPN/ov_export.py
pj3iL: Not fully resolvable from module structure, changed wherever possible, added documentations/error thrown.

allnes (approved):

pj3iL:
@ValentinaKats please let me know if any changes are required.

PR #1093, opened 2026-07-01
[NVIDIA] LLM ChatGPT2 support

redradist:
## What

Enables **GPT-2 / ChatGPT-2** (`gpt2`, 124M) text generation on the NVIDIA GPU plugin.

GPT-2's QKV projection uses `Split` / `VariadicSplit` with **integer constant** inputs. After the plugin's `i64 → i32` precision downcast those constants are `i32`, but the plugin read them with a fixed 64‑bit type, reading past the end of the buffer. This PR fixes the constant reads and adds regression tests.

## Changes

- **`Split` axis over-read** (`src/ops/split.cpp`): the axis constant was read with `get_data_ptr<int64_t>()`, which over-reads a narrower (`i32`) constant. Read it with `cast_vector<int64_t>()`, which converts from the constant's actual element type.
- **`VariadicSplit` out-of-bounds for the `-1` ("remaining") length** (`src/ops/variadic_split.{cpp,hpp}`): `buildSplitIndexHelper` indexed the raw `split_lengths` (which may contain `-1`), so the running offset decreased and the index ran one element past `split_lengths` → the kernel read out of bounds. Use the resolved per-output sizes (`axis_sizes_`) instead.
- **Regression tests** (`tests/.../single_layer_tests/split.cpp`, `variadic_split.cpp`): an `i32`-axis `Split` and a `VariadicSplit` with `-1` remaining lengths. Both fail without the fixes and pass with them.

## Verification

- The regression tests above (fail before / pass after).
- End-to-end: GPT-2 (fp16, stateful) generates coherent text on the NVIDIA device via OpenVINO GenAI.

## Note on PagedAttention (out of scope for this PR)

OpenVINO GenAI's `LLMPipeline` defaults to the **ContinuousBatching** backend, which converts `ScaledDotProductAttention` into **`PagedAttentionExtension`** (paged KV‑cache). The NVIDIA plugin does not implement `PagedAttentionExtension`, so that path fails with `... PagedAttentionExtension ... Is not found in OperationRegistry`.

GPT‑2 runs on the plugin via the **stateful / SDPA** path — e.g. `LLMPipeline(model, "NVIDIA", ATTENTION_BACKEND="SDPA")`:

| Path | Effort | Result |
| --- | --- | --- |
| **A** — `ATTENTION_BACKEND="SDPA"` (stateful) | none (works today) | GPT‑2 generates; no continuous batching |
| **B** — implement `PagedAttentionExtension` | large, separate feature | GenAI default / continuous batching / higher serving throughput |

`PagedAttentionExtension` support (path B) is tracked separately.

### DIFF
--- a/modules/nvidia_plugin/src/ops/split.cpp
+++ b/modules/nvidia_plugin/src/ops/split.cpp
@@ -46,7 +46,9 @@ SplitOp::SplitOp(const CreationContext& context,
     const auto element_type = input_element_type;
 
     auto& data_shape = splitOp->get_input_shape(0);
-    const int64_t axis = *axisNode->get_data_ptr<int64_t>();
+    // cast_vector handles any integer axis element type (e.g. GPT-2 emits an i32
+    // axis); get_data_ptr<int64_t> would over-read a narrower constant.
+    const int64_t axis = axisNode->cast_vector<int64_t>().at(0);
     OPENVINO_ASSERT(axis >= 0 && axis < data_shape.size(), "Node name: ", GetName());
     OPENVINO_ASSERT(data_shape[axis] % num_splits_ == 0, "Node name: ", GetName());
     const size_t split_step_size =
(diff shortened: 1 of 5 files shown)


p-durandin:
build_jenkins

p-durandin (approved):

PR #1121, opened 2026-08-03
[JAVA_API] Fix inverted double-free guard in Wrapper.close()

allnes:
### Fix an inverted double-free guard in `Wrapper.close()`

`Wrapper.close()` released the native object only on the **second** call:

```java
if (!closed.compareAndSet(false, true)) {   // false on the 1st close, true after
    delete(nativeObj);
}
```

`compareAndSet(false, true)` returns `true` on the first close (CAS succeeds),
so `!true` skips `delete()` — the object is leaked. A later close then takes the
branch and frees it, so a third call would double-free. Inverting the condition
makes `delete()` run once, on the first successful CAS.

### DIFF
--- a/modules/java_api/src/main/java/org/intel/openvino/Wrapper.java
+++ b/modules/java_api/src/main/java/org/intel/openvino/Wrapper.java
@@ -27,7 +27,7 @@ public class Wrapper implements AutoCloseable {
 
     @Override
     public void close() {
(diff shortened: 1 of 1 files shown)


allnes:
@aobolensk please review

aobolensk (approved):
LGTM

alvoron (approved):

PR #1124, opened 2026-08-03
[java_api] byte[] Tensor, PreProcessSteps normalization ops, indexed output getter + Android CheckJNI fix

allnes:
### Details:
Small additions to the OpenVINO Java bindings so inference can be driven from Kotlin/Java **without OpenCV**, plus one correctness fix that Android's CheckJNI exposes. Each has JNI + a unit test.

- **`Tensor(ElementType, int[] dims, byte[] data)`** — feed a `u8` buffer (e.g. an NHWC camera frame) directly; previously only a native pointer (`long`) or `float[]/int[]/long[]` were accepted, so a `u8` frame needed OpenCV's `Mat.dataAddr()` or a wasteful float copy. JNI `TensorByte` validates element type (`u8`/`i8`) and length.
- **`PreProcessSteps.scale(float|float[])`, `mean(float|float[])`, `convert_layout(Layout)`, `convert_element_type(ElementType)`** — previously only `resize` was exposed. Lets `/255` normalization and NHWC→NCHW be fused into the compiled model via `PrePostProcessor`, so callers pass raw `u8` pixels.
- **`InferRequest.get_output_tensor(int index)`** — read multi-output models (e.g. detection + mask prototypes) by index; the API only had the single-output getter and `get_tensor(String)`.
- **CheckJNI fix:** `ArrayList.add` (returns `boolean`) was called via `CallObjectMethod` in six JNI helpers. Desktop JVMs tolerate it; Android's CheckJNI **aborts the process** (`the return type of CallObjectMethodV does not match boolean java.util.ArrayList.add`) the moment a list-returning API is used (`Core.get_available_devices`, `CompiledModel.inputs/outputs`, `Model.outputs`, `Any.asList`). Switched to `CallBooleanMethod`.

Tests added/updated under `src/test/java/org/intel/openvino/`: `TensorTests`, `PrePostProcessorTests`, `InferRequestTests`.

Motivated by a new Android YOLO demo (submitted separately) that uses these APIs, but they are general and useful on their own.


allnes:
@aobolensk please review

modules/java_api/src/main/java/org/intel/openvino/Core.java
```diff
@@ -123,7 +123,7 @@ public CompiledModel compile_model(Model model, final String device) {
      *
      * @param model Model object acquired from {@link Core#read_model}.
      * @param device Name of a device to load a model to.
-     * @param properties properties – Map of pairs: (property name, property value) relevant only
+     * @param properties properties - Map of pairs: (property name, property value) relevant only
```
aobolensk: not needed

aobolensk (approved):
In general LGTM

modules/java_api/src/main/java/org/intel/openvino/Core.java
allnes: thanks fixed

alvoron (approved):

PR #1134, opened 2026-08-13
[genai_optimizations] Remove module

allnes:
- Remove the `genai_optimizations` module pending resolution of third-party licensing issues.
- Remove the corresponding Dependabot and labeling configuration.


allnes:
Closing this draft: the module must be retained and the affected implementations will be replaced instead.

